PR #8, opened 2024-01-21
got rid of junk

noahheller:


src/main/java/frc/robot/Constants.java
```diff
@@ -4,35 +4,14 @@
...
-    public static final boolean ARM_DEBUG = false;
-    public static final boolean EXTENDER_DEBUG = false;
-
-    public static final boolean GRIPPER_DEBUG = false;
-    public static final boolean PDB_DEBUG = false;
-    public static final boolean APRILTAG_DEBUG = false;
-
-    public static final boolean ENABLE_LOGGING = true;
+    public static class OperatorConstants {
+        public static final int kDriverControllerPort = 0;
```
bengold22: This is the pattern that comes with the WPILIB default project and isn't one we've adopted in the past (nested OperatorConstants inside Constants). We also don't use the variable style used in wpilib (k..). Can we just switch this to a flat uppercase constant?

bengold22 (approved):
Approved, one request for another constant adjustment and I do think we should move the code in RobotContainer into a Drivetrain subsystem, but that doesn't have to be this PR.

src/main/java/frc/robot/Constants.java
noahheller: done

PR #11, opened 2024-01-25
LV shooter code

WidestDiamond90:
Here is my shooter code with PID. There is a Shooter subsystem, a ShootTest command to spin for 2 seconds to test the motors, and a Shoot command to spin the motors until an optical sensor is activated.


src/main/java/frc/robot/Constants.java
```diff
@@ -5,6 +5,49 @@
...
+  public static final double SHOOTER_MOTOR_2_PID_P = 5e-5;
+  public static final double SHOOTER_MOTOR_2_PID_I = 0;
+  public static final double SHOOTER_MOTOR_2_PID_D = 5e-5;
+  public static final double SHOOTER_MOTOR_2_PID_IZ = 0;
+  public static final double SHOOTER_MOTOR_2_PID_FF = 0.000015;
+  public static final double SHOOTER_MOTOR_2_MAX_OUTPUT = 1;
+  public static final double SHOOTER_MOTOR_2_MIN_OUTPUT = -1;
+
+
+  public static class OperatorConstants {
```
batchen1: This should not be in the OperatorConstants class, take a look how we created the 2 joysticks.

src/main/java/frc/robot/Constants.java
```diff
@@ -5,6 +5,49 @@
...
+  public static final int SHOOTER_SENSOR_ID_1 = 0;
+  public static final int SHOOTER_SENSOR_ID_2 = 1;
+
+  public static final double SHOOTER_MOTOR_SPEED = 0.75;
+
+  public static final double SHOOTER_MOTOR_1_RPM = 12000;
+  public static final double SHOOTER_MOTOR_2_RPM = 12000;
+
+  //Shooter motor 1 PID constants
+  public static final double SHOOTER_MOTOR_1_PID_P = 5e-5;
```
batchen1: The PID values for both motors will be the same because it is the same type of motor and motor controller, so you can have a SHOOTER_MOTOR_PID_P (and all the other values) and use it for both motors. The only value that may be different between the 2 motors is the RPM.

src/main/java/frc/robot/Constants.java
```diff
@@ -5,6 +5,49 @@
...
+  public static final double SHOOTER_MOTOR_2_PID_IZ = 0;
+  public static final double SHOOTER_MOTOR_2_PID_FF = 0.000015;
+  public static final double SHOOTER_MOTOR_2_MAX_OUTPUT = 1;
+  public static final double SHOOTER_MOTOR_2_MIN_OUTPUT = -1;
+
+
+  public static class OperatorConstants {
+    public static final int kDriverControllerPort = 2;
+  }
+  public static final boolean DRIVETRAIN_DEBUG = false;
```
batchen1: I think this may be a conflict resolution mistake, I believe these were deleted.

PR #12, opened 2024-01-27
Sp ramp subsystem

Soham-West:


src/main/java/frc/robot/commands/RampMove.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public RampMove(Ramp ramp, double degrees) {
+        this.ramp = ramp;
+        this.degrees = degrees;
+        addRequirements(this.ramp);
+    }
+
+    @Override
+    public void initialize() {
+        super.initialize();
```
bengold22: You don't need to call super.initialize() here, the base class for Command doesn't do anything.

src/main/java/frc/robot/commands/RampMove.java
```diff
@@ -0,0 +1,36 @@
...
+        ramp.setRampPos(degrees);
+    }
+
+    @Override
```
bengold22: Since execute() and end() don't do anything, they don't need to be implemented.

src/main/java/frc/robot/subsystems/Ramp.java
```diff
@@ -0,0 +1,92 @@
...
+        encoder.setPosition(0);
+    }
+
+    public void setPID() {
+        pidController.setP(pidP);
+        pidController.setI(pidI);
+        pidController.setD(pidD);
+        pidController.setFF(pidFF);
+
+        //System.out.println("-------------->>>>>>> Set pid to " + pidP + "/" + pidI + "/" + pidD + "/" + pidFF);
```
bengold22: Remove unneeded commented out code

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -76,6 +84,12 @@ public RobotContainer() {
   private void configureBindings() {
     drivetrain.setDefaultCommand(new Drive(drivetrain, ()-> joyleft.getY(), ()-> joyleft.getX(), ()-> joyright.getX()));
   }
+  public void putShuffleboardCommands() {
+    controller.button(XboxController.Button.kA.value).onTrue(new RampMove(ramp, 400));
+    controller.button(XboxController.Button.kB.value).onTrue(new RampMove(ramp, 500));
+    SmartShuffleboard.putCommand("Ramp", "SetArmPID", new RampMove(ramp, 1200));
```
bengold22: If the command on the shuffleboard is working, convert this to be two commands on the shuffleboard: one that moves to 400, one that moves to 500, and remove the button mapping. This will be more scalable as more subsystems are added.

bengold22 (changes requested):
Can you pull main into this branch and resolve any conflicts?

PR #13, opened 2024-01-27
Lv shooter code

WidestDiamond90:
This is the shooter code with PID. There is 1 shooter subsystem, a ShootTest command to spin the motors for 2 seconds, and a second command called Shoot to spin the motors until an optical sensor has been activated.


src/main/java/frc/robot/Constants.java
```diff
@@ -5,6 +5,31 @@
...
+    public static final double SHOOTER_MOTOR_PID_P = 5e-5;
+    public static final double SHOOTER_MOTOR_PID_I = 0;
+    public static final double SHOOTER_MOTOR_PID_D = 5e-5;
+    public static final double SHOOTER_MOTOR_PID_IZ = 0;
+    public static final double SHOOTER_MOTOR_PID_FF = 0.000015;
+    public static final double SHOOTER_MOTOR_MAX_OUTPUT = 1;
+    public static final double SHOOTER_MOTOR_MIN_OUTPUT = -1;
+    
+    //Controller
+    public static final int kDriverControllerPort = 2;
```
bengold22: We don't use the "wpilib" style of variable naming. Our constants should be all caps with underscores.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -9,45 +9,62 @@
...
         configureBindings();
     }
 
     private void configureBindings() {
-        drivetrain.setDefaultCommand(new Drive(drivetrain, joyleft::getY, joyleft::getX, joyright::getX));
-    SmartShuffleboard.putCommand("TEST","WheelAlign",new WheelAlign(drivetrain));
-  }
 
+    //Create Shoot Button (A/B)
+    controller.button(XboxController.Button.kA.value).onTrue(new Shoot(shooter));
```
bengold22: I'm recommending we put these commands onto the shuffleboard as buttons instead of mapping to xbox controller buttons, that way you don't need a controller to test them and it reduces mapping dependencies. You can see on line 66 how to add it to the shuffleboard.

src/main/java/frc/robot/subsystems/Shooter.java
```diff
@@ -0,0 +1,134 @@
...
+  public Shooter() {
+    
+    this.shooterMotor1 = new CANSparkMax(Constants.SHOOTER_MOTOR_ID_1, CANSparkLowLevel.MotorType.kBrushless);
+    this.shooterMotor2 = new CANSparkMax(Constants.SHOOTER_MOTOR_ID_2, CANSparkLowLevel.MotorType.kBrushless);
+    this.shooterSensor1 = new DigitalInput(Constants.SHOOTER_SENSOR_ID_1);
+    this.shooterSensor2 = new DigitalInput(Constants.SHOOTER_SENSOR_ID_2);
+
+    shooterMotor1.restoreFactoryDefaults();
+    shooterMotor2.restoreFactoryDefaults();
+
```
batchen1, line 36: I think you need to invert one of the motors

batchen1 (changes requested):

batchen1 (approved):

PR #17, opened 2024-01-30
made naming work for red and blue

noahheller:


noahheller:
made new pull request

PR #26, opened 2024-02-06
added diagnostics to the drivetrain

Soham-West:


src/main/java/frc/robot/subsystems/swervev2/SwerveDrivetrain.java
```diff
@@ -75,7 +81,23 @@ public SwerveDrivetrain(SwerveIdConfig frontLeftConfig, SwerveIdConfig frontRigh
         this.frontRight.getSwerveMotor().getSteerMotor().setInverted(Constants.SWERVE_MODULE_PROFILE.isSteerInverted());
         this.backLeft.getSwerveMotor().getSteerMotor().setInverted(Constants.SWERVE_MODULE_PROFILE.isSteerInverted());
         this.backRight.getSwerveMotor().getSteerMotor().setInverted(Constants.SWERVE_MODULE_PROFILE.isSteerInverted());
-    }
+
+        Robot.getDiagnostics().addDiagnosable(new DiagSparkMaxEncoder("DT Drive", "Front Left", Constants.DIAG_REL_SPARK_ENCODER, (CANSparkMax) frontLeft.getSwerveMotor().getDriveMotor()));
```
bengold22, line 85: @armadilloz Diagnostics uses motor-specific diagnostics instead of a generic motor controller. I had Soham cast to CANSparkMax when we were writing this but now I wonder if there should be a generic diag class as well.

src/main/java/frc/robot/Constants.java
```diff
@@ -4,6 +4,6 @@
 
 package frc.robot;
 
-public final class Constants extends Constants2024{
+public final class Constants extends Constants2023{
```
bengold22: @Soham-West  we should generally probably put this back to Constants2024 when doing the PR back to main.

PR #27, opened 2024-02-06
added vector utils to calculate target angle for shooter

noahheller:


src/main/java/frc/robot/utils/math/VectorUtils.java
```diff
@@ -0,0 +1,46 @@
...
+        if (direct) return isDirect(velocityVector, deltaX) ? velocityVector : null;
+        else return velocityVector;
+    }
+
+    /**
+     * @param velocityVector the initial velocity and angle of the projectile
+     * @return the X displacement of the projectile following a theoretical parabola with level starting and ending y values
+     */
+    public static double horizontalRange(VelocityVector velocityVector) {
+        return ((Math.pow(velocityVector.getVelocity(), 2) * Math.sin(velocityVector.getAngle().getRadians() * 2))) / 9.8;
```
MatthewSpofford: If I'm understanding this right, I think the `9.8` is supposed to be `Constants.GRAVITY`?

noahheller: done

armadilloz (approved):
Approved: This will not break the code, but needs another review for the Math's correctness.

PR #36, opened 2024-02-11
while you hold trigger of left joystick your angle always faces speak…

noahheller:
…er, works for red and blue. Coords need to be adjusted, they are approximations :)


src/main/java/frc/robot/Alignable.java
```diff
@@ -0,0 +1,38 @@
+package frc.robot;
+
+import edu.wpi.first.math.geometry.Rotation2d;
+
+import java.util.function.BiFunction;
+
+public enum Alignable {
```
armadilloz, line 7: Lets separate the two concerns: One for the indication by the driver (where to aim) and the other for the drive calculations (angle and speed).
Have the enum be the location, and create a class (say, AutoAlignment) that does the calculation.

src/main/java/frc/robot/Alignable.java
```diff
@@ -0,0 +1,38 @@
...
+    AMP(0,0,(x,y)-> new Rotation2d(Math.PI / 2)),
```
armadilloz: I'm guessing from this that we have the AMP in a fixed angle (90 deg?). Should that not be 270 deg for RED?

src/main/java/frc/robot/Alignable.java
```diff
@@ -0,0 +1,38 @@
...
+    SPEAKER(0.46,5.67,(x2,y2)-> new Rotation2d(RobotContainer.shouldFlip() ? 0 : Math.PI).plus(new Rotation2d(Math.atan(y2/x2))));
```
armadilloz: Its probably time to rename `shouldFlip` with isRedAliance()

src/main/java/frc/robot/Alignable.java
```diff
@@ -0,0 +1,38 @@
...
+    private final BiFunction<Double, Double, Rotation2d> functionFromDist;
+
+    Alignable(double blueX, double y, BiFunction<Double,Double,Rotation2d> functionFromDist) {
+        this.blueX = blueX;
+        this.y = y;
+        this.functionFromDist = functionFromDist;
+    }
+
+    public double getX() {
+        return RobotContainer.shouldFlip() ? blueX + 16 : blueX;
```
armadilloz: what's the 16 for?

src/main/java/frc/robot/Alignable.java
armadilloz: I think you should "mirror image" the angle for red, not add 180. I'd like to see why this works.

src/main/java/frc/robot/Drive.java
```diff
@@ -1,27 +1,29 @@
...
 
-
-    public Drive(SwerveDrivetrain drivetrain, DoubleSupplier fwdSupplier, DoubleSupplier strSupplier, DoubleSupplier rtSupplier) {
-        addRequirements(drivetrain);
+    public Drive(SwerveDrivetrain drivetrain, DoubleSupplier fwdSupplier, DoubleSupplier strSupplier, DoubleSupplier rtSupplier, Supplier<Alignable> autoAlign) {
         this.drivetrain = drivetrain;
         this.fwdSupplier = fwdSupplier;
         this.strSupplier = strSupplier;
         this.rtSupplier = rtSupplier;
+        this.autoAlign = autoAlign;
```
armadilloz: probably `autoAim`

src/main/java/frc/robot/Drive.java
```diff
@@ -33,9 +35,10 @@ public void initialize() {
     public void execute() {
         double fwd = MathUtil.applyDeadband(fwdSupplier.getAsDouble()*Constants.MAX_VELOCITY,0.3);
         double str = MathUtil.applyDeadband(strSupplier.getAsDouble()*Constants.MAX_VELOCITY, 0.3);
-        double rcw = MathUtil.applyDeadband(rtSupplier.getAsDouble()*Constants.MAX_VELOCITY, 0.3);
-
-        ChassisSpeeds driveStates = drivetrain.createChassisSpeeds(fwd*(shouldFlip?1:-1), str*(shouldFlip?1:-1), -rcw, Constants.FIELD_RELATIVE);
+        double rcw;
+        if (autoAlign.get() == null) rcw = MathUtil.applyDeadband(rtSupplier.getAsDouble() * Constants.MAX_VELOCITY, 0.3);
```
armadilloz: Get the value from the provider once at the top, to reduce the risk of it changing between readings.

src/main/java/frc/robot/Alignable.java
```diff
@@ -0,0 +1,38 @@
...
+        return y;
+    }
+
+    public Rotation2d getAngle(double x, double y) {
+        return functionFromDist.apply(x - getX(), y - getY());
+    }
+    public double calcTurnSpeed(Rotation2d currentAngle, double x, double y){
+        double diff = currentAngle.minus(getAngle(x,y)).getDegrees();
+        if (Math.abs(diff) < 2) return 0;
+        return (Math.abs(diff)) / (360) * Math.signum(diff);
```
armadilloz: I think we need a proper PID here, with acceleration limit and maximum and minimum  speed

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -112,16 +114,22 @@ public void putShuffleboardCommands() {
 
         }
         if (Constants.FEEDER_DEBUG){
-            SmartShuffleboard.putCommand("Feeder", "Feed", new StartFeeder(feeder));
+//            SmartShuffleboard.putCommand("Feeder", "Feed", new StartFeeder(feeder));
```
armadilloz: ?

armadilloz (changes requested):

src/main/java/frc/robot/Alignable.java
noahheller: amp is on the same side for both teams

src/main/java/frc/robot/Alignable.java
noahheller: my bad, I fixed this

src/main/java/frc/robot/Alignable.java
noahheller: done

src/main/java/frc/robot/Alignable.java
noahheller: done

src/main/java/frc/robot/Alignable.java
noahheller: for red the robot faces the "0" degree when looking at speaker headon, for blue the robot faces "180" degrees when looking head on. the math works for red but you need to add the 180 for blue to accommodate for the offset. Furthermore, using Rotation2d.add() keeps the range between -PI/2 and PI/2, necessary for the math to work

src/main/java/frc/robot/Drive.java
noahheller: good call, done

src/main/java/frc/robot/RobotContainer.java
noahheller: feeder causes issues on the 2023 robot. code immedatly crashes.

src/main/java/frc/robot/Alignable.java
```diff
@@ -5,21 +5,23 @@
 import java.util.function.BiFunction;
 
 public enum Alignable {
-    AMP(0,0,(x,y)-> new Rotation2d(Math.PI / 2)),
-    SPEAKER(0.46,5.67,(x2,y2)-> new Rotation2d(RobotContainer.shouldFlip() ? 0 : Math.PI).plus(new Rotation2d(Math.atan(y2/x2))));
+    AMP(0,0,0,(x,y)-> new Rotation2d(Math.PI / 2)),
+    SPEAKER(0.46,16.46,5.67,(x2,y2)-> new Rotation2d(RobotContainer.isRedAlliance() ? 0 : Math.PI).plus(new Rotation2d(Math.atan(y2/x2))));
 
     private final double blueX;
+    private final double redX;
```
armadilloz, line 12: What is redX and why do we need two x coords?

src/main/java/frc/robot/Alignable.java
```diff
@@ -29,10 +31,4 @@ public double getY() {
     public Rotation2d getAngle(double x, double y) {
```
armadilloz: getAngle needs to be moved to the AutoAlignment class as well. The point of this change was to separate the various field elements from the driving, and getAngle(), with its field location calculation, is driving instructions. You will end up having a case statement in there for the various positions, which affords better readability of what is going on.

armadilloz (changes requested):

src/main/java/frc/robot/Alignable.java
armadilloz: I still don't see it. True that in case of 0 (Bobot in centerline of field) the angle is 180 inverted for blue/red, but in non-0 locations the difference is not 180. If you move "right" you will have >0 angle towards red and <180 angle towards Blue

src/main/java/frc/robot/RobotContainer.java
armadilloz: Is Main broken? This should be fixed - not disabled. You should be able to test locally like that but not commit back.

src/main/java/frc/robot/Alignable.java
noahheller: I got confused myself. I am going to write some unit tests a figure out what is going on

src/main/java/frc/robot/Alignable.java
noahheller: I decided to make a separate red and blue x, instead of applying an offset like I did in FieldLocation because I thought there might be a situation where you want to a align with something that is not mirrored on the field. Upon further thought I cant think of such as situation. LMK if I should apply an offset instead

src/main/java/frc/robot/Alignable.java
noahheller: my bad I don't know why I did not copy this method over as well

src/main/java/frc/robot/Alignable.java
noahheller: I think a pid would be a good idea, however I feel like that could be done at a latter date and want to get a proof of concept going. However, if you want to do the PID now I understand that aswell

armadilloz: That's a good candidate for deferral.

src/main/java/frc/robot/Alignable.java
armadilloz: That is fine. Once I understand what it is for, it makes sense. We may need redY once we have the full list of items to align on, but this is good for now.

src/main/java/frc/robot/Alignable.java
```diff
@@ -6,7 +6,7 @@
 
 public enum Alignable {
     AMP(0,0,0,(x,y)-> new Rotation2d(Math.PI / 2)),
-    SPEAKER(0.46,16.46,5.67,(x2,y2)-> new Rotation2d(RobotContainer.isRedAlliance() ? 0 : Math.PI).plus(new Rotation2d(Math.atan(y2/x2))));
+    SPEAKER(0.46,16.46,5.67,(x2,y2)->new Rotation2d(Math.atan(y2/x2)));
```
armadilloz, line 9: Shouldn't redX be (16-0.46) since it's closer to Blue than the end of the field?

armadilloz (approved):
Approving as-is, but there are a few topics that we should discuss:
- redX send to be outside the field
- I believe the angle calculation is still unclear - it should be tan(y/x) for Red and 180-atan(y/x) for Blue according to my understanding.
- Question about the button assignment: if you press both then release one, the subsystem will have null though one button is still pressed. Not a big issue, though, if drivers are OK
- Some cleanup

PR #37, opened 2024-02-11
added servo

Levercpu:
added servo functionality


src/main/java/frc/robot/subsystems/ServoSubsystem.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+  public void set_Servo(double position) {
+    servo.set(position);
+  }
+  public double get_Servo() {
+    return servoPosition;
+  }
+  public double get_Speed() {
+    return servoSpeed;
+  }
```
MatthewSpofford: Instead of these getter methods returning variables that were set in `periodic()`, these methods should just be directly returning the values of `servo.getPosition()`/`servo.getSpeed()`. Then for the climber debug values being set during periodic, they can call the getters instead.

src/main/java/frc/robot/commands/RaiseArms.java
```diff
@@ -0,0 +1,43 @@
...
+  @Override
+  public void end(boolean interrupted) {}
+
+  // Returns true when the command should end.
+  @Override
+  public boolean isFinished() {
+    if (Timer.getFPGATimestamp() - startTime > Constants.RAISING_TIMEOUT) {
+      return true;
+    }
+    return false;
```
MatthewSpofford: Could be simpler written like so:
```suggestion
    return Timer.getFPGATimestamp() - startTime > Constants.RAISING_TIMEOUT;
```

src/main/java/frc/robot/subsystems/ServoSubsystem.java
MatthewSpofford: Proper Java naming convention is using camelCase, which would not use an "_" in the name. Instead, you should rename these methods to something like `getServo`, or `setServo`.

src/main/java/frc/robot/commands/RatchetingMechanism.java
```diff
@@ -0,0 +1,40 @@
...
+    addRequirements(servo);
+  }
+
+  // Called when the command is initially scheduled.
+  @Override
+  public void initialize() {
+    if (servo.get_Servo() == 0.0) {
+      servo.set_Servo(1.0);
+    } else if (servo.get_Servo() == 1.0) {
+      servo.set_Servo(0.0);
```
MatthewSpofford: These servo positions should be written in the `Constants` class.

As this logic is written, this means this command will do nothing when the servo is not at positions 0.0 or 1.0. It may be useful to also handle the conditions where the servo is between 0 and 1.

src/main/java/frc/robot/commands/RaiseArms.java
```diff
@@ -0,0 +1,43 @@
...
+    this.climber = climber;
+    addRequirements(climber);
+  }
+  private double startTime;
+
+  // Called when the command is initially scheduled.
+  @Override
+  public void initialize() {
+    startTime = Timer.getFPGATimestamp();
+    climber.raise(true);
```
MatthewSpofford: This `raise()` call should be moved to `execute()`, since the motor will need to be moving for the entire time this command is being executed

MatthewSpofford (changes requested):
Needs some additional changes to the climber commands, and servo command/subsystem

build.gradle
```diff
@@ -1,6 +1,6 @@
 plugins {
     id "java"
-    id "edu.wpi.first.GradleRIO" version "2024.1.1"
+    id "edu.wpi.first.GradleRIO" version "2024.2.1"
```
MatthewSpofford: Need to change back to 2024.1.1 for now

src/main/java/frc/robot/Constants2024.java
```diff
@@ -92,12 +92,15 @@ public class Constants2024 extends GameConstants {
     public static final int OUTTAKE_MOTOR2_ID = 7;
     public static final int CLIMBER_MOTOR1_ID = 44;
     public static final int CLIMBER_MOTOR2_ID = 45;
+    public static final int SERVO_ID = 0;
```
noahheller: all constants added to 2024 MUST be added to 2023 or the code will not build

src/main/java/frc/robot/subsystems/ServoSubsystem.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class ServoSubsystem extends SubsystemBase {
+  private final Servo servo;
+  private double servoPosition;
+  private double servoSpeed;
+  /** Creates a new Servo. */
+  public ServoSubsystem() {
+    this.servo = new Servo(Constants.SERVO_ID);
+  }
+  public void set_Servo(double position) {
```
bengold22: I played with the servo APIs, there's a "setAngle" that I believe we want to use, not .set(position).

src/main/java/frc/robot/subsystems/ServoSubsystem.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+  public double get_Servo() {
+    return servoPosition;
+  }
+  public double get_Speed() {
+    return servoSpeed;
+  }
+  @Override
+  public void periodic() {
+    servoPosition=servo.getPosition();
```
bengold22: Anything in periodic should be wrapped in an "if (DEBUG)" statement so it can be disabled under normal running operation.

bengold22 (changes requested):
Several comments left to address on the PR

src/main/java/frc/robot/Constants2023.java
```diff
@@ -12,6 +12,11 @@ public class Constants2023 extends GameConstants {
     public static final double RAMP_MAX_RPM_VELOCITY = 500;
     public static final double RAMP_MAX_RPM_ACCELERATION = 1500;
 
+    //Servo
+    public static final int SERVO_POS_0 = 0;
+    public static final int SERVO_POS_1 = 180;
+    public static final int SERVO_ID = 0;
```
bengold22: If this is declared in one constant it should be in all.

src/main/java/frc/robot/commands/RatchetingMechanism.java
bengold22, line 1: I don't think this is how we want the command to work. Toggling doesn't work the first time because getPosition() always returns 0 regardless of size. What we need is a specific "set servo to angle" method. We'll determine the final target angles by experimenting with the real robot.

src/main/java/frc/robot/subsystems/ServoSubsystem.java
bengold22, line 1: We were talking about this and we want to move the servos to be part of the climber. There will be two on the real robot and should move together.

bengold22 (changes requested):

src/main/java/frc/robot/constants/Constants.java
```diff
@@ -5,4 +5,7 @@
 package frc.robot.constants;
 
 public final class Constants extends Constants2024{
+
```
armadilloz: Extra spaces?

src/main/java/frc/robot/subsystems/Climber.java
```diff
@@ -2,19 +2,22 @@
...
     private final CANSparkMax SparkMax2;
+    private final Servo servo;
     private final AHRS navxGyro;
 
-    public Climber(AHRS navxGyro){
+    public Climber(AHRS navxGyro, Servo servo) {
         this.SparkMax1 = new CANSparkMax(Constants.CLIMBER_MOTOR1_ID, CANSparkMax.MotorType.kBrushless);
         this.SparkMax2 = new CANSparkMax(Constants.CLIMBER_MOTOR2_ID, CANSparkMax.MotorType.kBrushless);
         this.navxGyro = navxGyro;
+        this.servo = servo;
```
armadilloz: Just use the constants here instead of passing as a parameter (You'd need two of them)

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -111,7 +113,7 @@ private void setupDriveTrain() {
         KinematicsConversionConfig kinematicsConversionConfig = new KinematicsConversionConfig(Constants.WHEEL_RADIUS, Constants.SWERVE_MODULE_PROFILE.getDriveRatio(), Constants.SWERVE_MODULE_PROFILE.getSteerRatio());
         SwervePidConfig pidConfig = new SwervePidConfig(drivePid, steerPid, driveGain, steerGain, constraints);
         AHRS navxGyro = new AHRS();
-        climber = new Climber(navxGyro);
+        climber = new Climber(navxGyro, new Servo(Constants.SERVO_ID));
```
armadilloz: We're going to need two of them, and no need for params)

src/main/java/frc/robot/commands/RaiseArms.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  // Called once the command ends or is interrupted.
+  @Override
+  public void end(boolean interrupted) {}
+
+  // Returns true when the command should end.
+  @Override
+  public boolean isFinished() {
+    return Timer.getFPGATimestamp() - startTime > Constants.RAISING_TIMEOUT;
```
armadilloz, line 40: Add todo: This needs to be considering the limit switches.

src/main/java/frc/robot/commands/climber/SetServoAngle.java
```diff
@@ -0,0 +1,25 @@
+package frc.robot.commands.climber;
+
+import edu.wpi.first.wpilibj2.command.Command;
+import frc.robot.subsystems.Climber;
+
+public class SetServoAngle extends Command {
+    private final Climber climber;
+    private final double targetAngle;
```
armadilloz: need two angels (right and left - they may be different).

src/main/java/frc/robot/constants/Constants2023.java
```diff
@@ -14,6 +14,11 @@ public class Constants2023 extends GameConstants {
     public static final double RAMP_MAX_RPM_VELOCITY = 500;
     public static final double RAMP_MAX_RPM_ACCELERATION = 1500;
 
+    //Servo
+    public static final int SERVO_POS_0 = 0;
```
armadilloz: Need two servos

src/main/java/frc/robot/constants/Constants2024.java
```diff
@@ -15,6 +15,11 @@ public class Constants2024 extends GameConstants {
     public static final double RAMP_MAX_RPM_VELOCITY = 500;
     public static final double RAMP_MAX_RPM_ACCELERATION = 1500;
 
+    //Servo
+    public static final int SERVO_POS_0 = 0;
```
armadilloz: Need two servos

src/main/java/frc/robot/subsystems/Climber.java
```diff
@@ -26,16 +29,23 @@ public double getGyroRoll() {
         return (navxGyro.getRoll() % 360);
     }
 
+    /**
+     * @param angle in degrees
+     */
+    public void setServoAngle(double angle){
```
armadilloz: needs two angles

armadilloz (changes requested):

armadilloz (approved):

PR #40, opened 2024-02-13
Added a command to set the ramp to the top at the beginning of the match

neo-Aditya:
Tested and works


src/main/java/frc/robot/subsystems/Ramp.java
```diff
@@ -74,6 +74,14 @@ public double getRampPos() {
         return encoder.getPosition();
     }
 
+    public boolean getForwardSwitch() {
```
noahheller: Method name is unclear (what does true/false mean, ie pressed or not). I would recommend adding Javadocs and/or renaming method

src/main/java/frc/robot/commands/ResetRamp.java
```diff
@@ -0,0 +1,44 @@
...
+  // Called when the command is initially scheduled.
+  @Override
+  public void initialize() {
+    startTime = Timer.getFPGATimestamp();
+  }
+
+  // Called every time the scheduler runs while the command is scheduled.
+  @Override
+  public void execute() {
+    ramp.changeRampPos(99);
```
bengold22: This seems like a strange way to do this. First, how do you know which way is "forward"? Should this 99 be a constant? Why 99? Why not 9999?

src/main/java/frc/robot/commands/ResetRamp.java
bengold22, line 1: Leave a TODO in this code about how it actually needs to be tested, so we'll know what to do when we get the actual robot.

src/main/java/frc/robot/Robot.java
```diff
@@ -62,6 +63,7 @@ public void autonomousInit() {
     @Override
     public void autonomousPeriodic() {
         loopTime = Timer.getFPGATimestamp();
+        new ResetRamp(robotContainer.getRamp());
```
bengold22: Putting this in autonomousPeriodic does not seem correct. You would be rescheduling this over and over, 50 times a second.

src/main/java/frc/robot/commands/ResetRamp.java
neo-Aditya: Just added it.

src/main/java/frc/robot/Robot.java
neo-Aditya: Moved it to AutonomousInit

src/main/java/frc/robot/commands/ResetRamp.java
neo-Aditya: we're using motor speed now, still assuming a lot

bengold22 (approved):

src/main/java/frc/robot/commands/ResetRamp.java
```diff
@@ -22,23 +29,28 @@ public ResetRamp(Ramp ramp) {
   @Override
   public void initialize() {
     startTime = Timer.getFPGATimestamp();
+    ramp.setMotor(0.3); //assuming positive is forward with a random speed
```
noahheller: make this a constants

neo-Aditya: Although I don't see a use for this when we don't have it refined yet, I added it

PR #41, opened 2024-02-13
MoveDistance in meters

neo-Aditya:
Doesn't move


src/main/java/frc/robot/commands/MoveDistance.java
```diff
@@ -0,0 +1,70 @@
...
+  // Called once the command ends or is interrupted.
+  @Override
+  public void end(boolean interrupted) {
+    drivetrain.stopMotor();
+  }
+
+  // Returns true when the command should end.
+  @Override
+  public boolean isFinished() {
+    if (Math.abs(startPoseX - drivetrain.getPose().getX()) > Math.abs(changeX) && Math.abs(startPoseY - drivetrain.getPose().getY()) > Math.abs(changeY)) {
```
noahheller: you might want to add a threshold to the distance between current and desired position

src/main/java/frc/robot/commands/drive/MoveDistance.java
bengold22, line 1: When is this command expected to be used?

neo-Aditya: I believe we may use it in auto, but I don't see where else

src/main/java/frc/robot/commands/drive/MoveDistance.java
```diff
@@ -0,0 +1,74 @@
...
+    drivetrain.stopMotor();
+  }
+
+  // Returns true when the command should end.
+  @Override
+  public boolean isFinished() {
+    if (Math.abs(startPoseX - drivetrain.getPose().getX()) > Math.abs(changeX) && Math.abs(startPoseY - drivetrain.getPose().getY()) > Math.abs(changeY)) {
+      return true;
+    }
+    return (((Timer.getFPGATimestamp() - startTime) >= 2) || underneathThreshold);
```
noahheller: This threshold does not do anything because it only gets update once in the constructor of the Command.

neo-Aditya: Sorry about not making it clear, the Threshhold is to make sure that the values given to the command is reasonable.

neo-Aditya: Whoops

PR #43, opened 2024-02-14
update wpi

noahheller:
This is NOT urgent but should be done at some point


bengold22 (approved):

bengold22:
I think sooner rather than later, better now than nearer to comps

PR #45, opened 2024-02-15
Intake

WidestDiamond90:
There is on Intake subsystem, and a command called StartIntake that runs the intake motors for 5 seconds. There is also a motor current limit in which the values still need to be tweaked.


.OutlineViewer/outlineviewer.json
bengold22, line 1: What is this file for? I'm not sure it should be checked in

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,58 @@
...
+        this.intakeMotor2 = new WPI_TalonSRX(Constants.INTAKE_MOTOR_2_ID);
+        
+        intakeMotor1.setInverted(false);
+        intakeMotor2.setInverted(false);
+
+        //Configure current limits *VALUES HAVE TO BE TWEAKED*
+        intakeMotor1.configPeakCurrentLimit(Constants.INTAKE_MOTOR_PEAK_CURRENT_LIMIT);
+        intakeMotor1.configPeakCurrentDuration(Constants.INTAKE_MOTOR_PEAK_CURRENT_DURATION);
+        intakeMotor1.configContinuousCurrentLimit(Constants.INTAKE_MOTOR_CONTINUOUS_CURRENT_LIMIT);
+        intakeMotor1.enableCurrentLimit(true);
```
bengold22: Add a constant for whether the current limit should be enabled, and start with it false. We'll have to find the right numbers and if we run with current limiting to begin with it will give us incorrect behavior.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public void stopMotors() {
+        intakeMotor1.set(0);
+        intakeMotor2.set(0);
+    }
+
+    @Override
+    public void periodic() {
+        SmartShuffleboard.put("Intake", "Intake Motor 1 Speed", getMotor1Speed());
```
bengold22, line 53: Create a new boolean in GameConstants, INTAKE_DEBUG, and wrap everything in period in a check to see if INTAKE_DEBUG is true.

if (constants.INTAKE_DEBUG) {
  SmartShuffleboard.put(..)

bengold22 (approved):

PR #47, opened 2024-02-15
Nh package organize

noahheller:
Hopefully this does not break anything with git. This is not important and provides functional change besides making the code easier to find.


bengold22:
If this is going to be merged it better be soon, this is going to cause merge conflicts with everyone that has started on work from the existing code.

batchen1 (approved):

PR #49, opened 2024-02-17
A command to shoot from feeder

neo-Aditya:
Tested, timeout's being weird though

### DIFF
--- a/src/main/java/frc/robot/commands/cannon/ShootPiece.java
+++ b/src/main/java/frc/robot/commands/cannon/ShootPiece.java
@@ -1,0 +1,21 @@
+// Copyright (c) FIRST and other WPILib contributors.
+// Open Source Software; you can modify and/or share it under the terms of
+// the WPILib BSD license file in the root directory of this project.
+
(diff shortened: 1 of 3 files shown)


src/main/java/frc/robot/commands/cannon/ShootPiece.java
```diff
@@ -0,0 +1,24 @@
...
+
+// NOTE:  Consider using this command inline, rather than writing a subclass.  For more
+// information, see:
+// https://docs.wpilib.org/en/stable/docs/software/commandbased/convenience-features.html
+public class ShootPiece extends ParallelRaceGroup {
+  /** Creates a new ShootPiece. */
+  public ShootPiece(Shooter shooter, Feeder feeder) {
+    addCommands(
+    new Shoot(shooter),
+    new SequentialCommandGroup(new StartFeeder(feeder), new WaitCommand(2))
```
bengold22: We probably want to wait just a bit before running the feeder, like .5 or less. Can the command do that?

bengold22 (changes requested):

PR #64, opened 2024-02-20
Xbox controller is able to replace the joysticks

neo-Aditya:


src/main/java/frc/robot/constants/GameConstants.java
```diff
@@ -7,8 +7,10 @@ public class GameConstants {
     public static final int LEFT_JOYSICK_ID = 0;
     public static final int RIGHT_JOYSTICK_ID = 1;
     public static final int XBOX_CONTROLLER_ID = 3;
+    public static final double XBOX_CONTROLLER_JOYSTICK_DEADZONE = 0.3;
     public static final double DIAG_REL_SPARK_ENCODER = 0.1;
     public static final int DIAG_ABS_SPARK_ENCODER = 20;
+    public static final boolean XBOX_JOYSTICK_DRIVE = false;
```
neo-Aditya: Not a great variable name

MatthewSpofford (changes requested):
Everything looks good, but for the Drive command I would recommend trying to get the `MathUtil.applyDeadband()` to work properly with the Xbox controller. If you can't, I would create a helper method for setting the drivetrain speeds so that we can reduce code reuse.

armadilloz:
Is this still needed?

PR #75, opened 2024-02-24
Bg climber changes

bengold22:
Redesigned climber code based on testing:

* Changed "left" and "right" to match a software-first view (battery is front)
* Simplified to a full manual control only (drive to limit switches)
* Added ratchet controls

Tested on 2024 robot, including climbing


src/main/java/frc/robot/constants/GameConstants.java
```diff
@@ -18,7 +18,7 @@ public class GameConstants {
     public static final boolean RAMP_PID_DEBUG = false;
     public static final boolean SWERVE_DEBUG = false;
     public static final boolean FEEDER_DEBUG = false;
-    public static final boolean CLIMBER_DEBUG = false;
+    public static final boolean CLIMBER_DEBUG = true;
```
armadilloz: please leave at false

src/main/java/frc/robot/subsystems/Climber.java
```diff
@@ -44,51 +50,41 @@ public void setLeftServoAngle(double angle){
...
-    }
-    public void balanceLeft(double speed) {
-        climberLeft.set(0);
-        climberRight.set(speed);
-    }
-    public void stop(){
-        climberLeft.set(0);
-        climberRight.set(0);
+
+    public void resetEncoders() {
```
armadilloz, line 73: Should we have a command for that just in case?

src/main/java/frc/robot/subsystems/Climber.java
```diff
@@ -44,51 +50,41 @@ public void setLeftServoAngle(double angle){
...
+        this.climberLeft.getEncoder().setPosition(0);
+        this.climberRight.getEncoder().setPosition(0);
+    }    
+
+    public void engageRatchet() {
```
armadilloz, line 78: Is this being used?

armadilloz: Maybe using these methods is better than setting the angles externally through the command?

armadilloz (changes requested):

src/main/java/frc/robot/subsystems/Climber.java
bengold22: Agreed, done

armadilloz (approved):

PR #76, opened 2024-02-24
Added Automatic Shooting Speeds based on gyro

Sahiltheram:


src/main/java/frc/robot/commands/shooter/ShootSpeaker.java
```diff
@@ -4,24 +4,25 @@
...
         addRequirements(shooter);
     }
 
     @Override
     public void initialize() {
         timer.reset();
         activated = true;
-        shooter.setShooterMotorLeftSpeed(Constants.SHOOTER_MOTOR_LEFT_SPEED);
-        shooter.setShooterMotorRightSpeed(Constants.SHOOTER_MOTOR_RIGHT_SPEED);
+        shooter.setShooterSpeeds(drivetrain);
```
armadilloz: Move the logic from the subsystem to the command - the subsystem should only be concerned with setting speed to the motors.

armadilloz: Commands for setting speed on the motors have to be executed every cycle or else the watchdog is going to disable the robot.

armadilloz (changes requested):

src/main/java/frc/robot/commands/shooter/ShootSpeaker.java
Sahiltheram: Done

src/main/java/frc/robot/commands/shooter/ShootSpeaker.java
```diff
@@ -31,6 +32,19 @@ public boolean isFinished() {
         return (timer.hasElapsed(Constants.SHOOTER_TIME_AFTER_TRIGGER)) && (activated);
     }
 
+    @Override
+    public void execute() {
+        if (((RobotContainer.isRedAlliance() == true) && (((((drivetrain.getGyroAngle().getDegrees()) % 360) + 360) % 360) > 180)) || ((RobotContainer.isRedAlliance() == false) && (((((drivetrain.getGyroAngle().getDegrees()) % 360) + 360) % 360) < 180))) {
```
bengold22: Why do you need to add 360 and then mod 360 again? Is it because it could be negative degrees and it's being ensured it's positive?

bengold22: OK I just threw this into Excel and as far as I can tell, +360 then %360 makes no difference to the number. EDIT: I think this is difference between Excel mod and java mod, Java will keep sign of numerator and Excel keeps sign of denominator.

bengold22 (approved):

src/main/java/frc/robot/commands/shooter/ShootSpeaker.java
```diff
@@ -32,6 +32,19 @@ public boolean isFinished() {
...
+            shooter.setShooterMotorRightSpeed(Constants.SHOOTER_MOTOR_LEFT_SPEED);
+            shooter.setShooterMotorLeftSpeed(Constants.SHOOTER_MOTOR_RIGHT_SPEED);
+        }
+        if (((RobotContainer.isRedAlliance() == false) && (((((drivetrain.getGyroAngle().getDegrees()) % 360) + 360) % 360) > 180)) || ((RobotContainer.isRedAlliance() == true) && (((((drivetrain.getGyroAngle().getDegrees()) % 360) + 360) % 360) < 180))) {
```
armadilloz: What if the angle is exactly 180?

src/main/java/frc/robot/commands/shooter/ShootSpeaker.java
```diff
@@ -32,6 +32,19 @@ public boolean isFinished() {
         return (timer.hasElapsed(Constants.SHOOTER_TIME_AFTER_TRIGGER)) && (activated);
     }
 
+    @Override
+    public void execute() {
+        if (((RobotContainer.isRedAlliance() == true) && (((((drivetrain.getGyroAngle().getDegrees()) % 360) + 360) % 360) > 180)) || ((RobotContainer.isRedAlliance() == false) && (((((drivetrain.getGyroAngle().getDegrees()) % 360) + 360) % 360) < 180))) {
+            shooter.setShooterMotorRightSpeed(Constants.SHOOTER_MOTOR_LEFT_SPEED);
```
armadilloz: Right and left are no longer accurate - that could change to high and low speeds.

armadilloz (changes requested):

src/main/java/frc/robot/commands/shooter/ShootSpeaker.java
armadilloz: Hint - this can be simplified with an `else` clause

src/main/java/frc/robot/commands/shooter/ShootSpeaker.java
Sahiltheram: Done

src/main/java/frc/robot/commands/shooter/ShootSpeaker.java
Sahiltheram: Done

armadilloz (approved):

PR #83, opened 2024-02-24
Made the ramp not go back to the previous position when teleop is enabled

Sahiltheram:


src/main/java/frc/robot/commands/feeder/FeederGamepieceUntilLeave.java
bengold22, line 1: Why is this PR modifying this file? I don't think it should be.

src/main/java/frc/robot/commands/sequences/ExitAndShoot.java
bengold22, line 1: Same here, why is this PR modifying the ExitAndShoot command?

src/main/java/frc/robot/constants/Constants2023.java
```diff
@@ -32,6 +32,9 @@ public class Constants2023 extends GameConstants {
     public static final double SHOOTER_MOTOR_LOW_SPEED = 0.6;
     public static final double SHOOTER_MOTOR_HIGH_SPEED = 1;
 
+    public static final double SHOOTER_MOTOR_RIGHT_AMP_SPEED = 0.25;
```
bengold22: These constants are good examples of values that can go into GameConstants so you don't have to repeat them in every constants file.

src/main/java/frc/robot/constants/Constants2023.java
```diff
@@ -32,6 +32,9 @@ public class Constants2023 extends GameConstants {
...
+    public static final double SHOOTER_MOTOR_LEFT_AMP_SPEED = 0.25;
```
bengold22: Also I don't think any of this should have been in this PR.

bengold22 (changes requested):

src/main/java/frc/robot/commands/feeder/FeederGamepieceUntilLeave.java
Sahiltheram: I was asked to make this small change and I didn't want to create a different PR

src/main/java/frc/robot/constants/Constants2023.java
Sahiltheram: Deleted

src/main/java/frc/robot/commands/sequences/ExitAndShoot.java
Sahiltheram: Fixed

PR #87, opened 2024-02-25
Fixed color sensor

neo-Aditya:
NEEDS TO BE TESTED


src/main/java/frc/robot/utils/ColorSensor.java
```diff
@@ -44,8 +44,7 @@ public ColorValue getColor() {
         return ColorValue.getFromColor(match.color);
     }
 
-    public ColorMatchResult getRawColor() {
-        Color detectedColor = colorSensor.getColor();
-        return colorMatcher.matchClosestColor(detectedColor);
+    public Color getRawColor() {
```
bengold22, line 50: Before we merge this, did you verify this is only used for logging purposes? It seems likely given that you changed the signature of the method but no other code had to change.

neo-Aditya: I just made another method that does what the old one did since I forgot that probability isn't part of color

src/main/java/frc/robot/subsystems/Feeder.java
```diff
@@ -57,13 +58,14 @@ public boolean pieceNotSeen() {
     public void periodic() {
         if (Constants.FEEDER_DEBUG) {
             ColorValue detectedColor = colorSensor.getColor();
-            ColorMatchResult rawColor = colorSensor.getRawColor();
+            Color rawColor = colorSensor.getRawColor();
+            ColorMatchResult matchedColor = colorSensor.getMatchedColor();
```
neo-Aditya, line 60: We probably need to test this before it gets merged

src/main/java/frc/robot/utils/ColorSensor.java
bengold22: What's the difference between "getColor()" and "getMatchedColor"? Instead of creating different public methods can you add one public method that exposes what people should be using?

neo-Aditya: As far as I can tell getColor gives a ColorValue that's used to get the enum with the name in it whilst getMatched Color gives a ColorMatchResult that's used in the feeder to get the probability of the sensor being right.

src/main/java/frc/robot/subsystems/Feeder.java
```diff
@@ -46,24 +47,23 @@ public ColorValue getColor() {
     }
 
     public boolean pieceSeen() {
-        return (getColor() == ColorValue.Piece);
-    }
-
-    public boolean pieceNotSeen() {
-        return (getColor() == ColorValue.Plastic);
+        ColorMatchResult latestResult = colorSensor.getMatchedColor();
+        return latestResult.confidence > Constants.COLOR_CONFIDENCE_RATE;
```
noahheller: you should set the confindence rate in the color macher itself (there is a setConfidenceThreshold method in colorMatch

src/main/java/frc/robot/utils/ColorSensor.java
```diff
@@ -44,8 +42,12 @@ public ColorValue getColor() {
         return ColorValue.getFromColor(match.color);
     }
 
-    public ColorMatchResult getRawColor() {
+    public ColorMatchResult getMatchedColor() {
```
noahheller, line 45: create a wrapper that containers the TargetColorResult (ColorValue enum) and a calucated confidence when getting the matched color. then made get Color return the Wrapped value.

src/main/java/frc/robot/utils/ColorSensor.java
```diff
@@ -44,8 +42,12 @@ public ColorValue getColor() {
...
         Color detectedColor = colorSensor.getColor();
```
noahheller, line 46: Replace first Two lines of getColor() with getMatchedColor()

bengold22 (approved):

PR #88, opened 2024-02-25
made auto align spin shooter motors with spin

noahheller:
tested


src/main/java/frc/robot/commands/shooter/AdvancedSpinningShot.java
```diff
@@ -0,0 +1,54 @@
...
+        shooter.stopShooter();
+        timer.stop();
+    }
+
+    @Override
+    public boolean isFinished() {
+        return timer.hasElapsed(Constants.ADVANCED_SPINNING_SHOT_TIMEOUT);
+    }
+
+    private ShooterSpeed calcuateShooterSpeed() {
```
bengold22, line 55: Is this trying to flip the motor speeds based on what side of the speaker we're on? @Sahiltheram  has done similar work. Can you two work together to determine what is needed?

noahheller: This code does a slightly different thing and will be used if we can get a higher level of automation. This code mitigates the shooting spin oscillation problem and should only be used in tangent with the auto steer code. Bat-Chen, Ohad, and I discussed this yesterday, and I would be happy to give you a more in-depth explanation in person.

MatthewSpofford (approved):

PR #99, opened 2024-02-25
Making ramp move till it's desired position

### LINKED ISSUE #97: Convert RampMove to not finish until ramp is in the right position
We want to use RampMove to be "synchronous" so the "isFinished" method should not return true until the reported encoder position is within some range of what the target PID position is.

If we want an asynchronous version (where we sit the PID value and immediately return that it is finished) I think we should create another command.

neo-Aditya:
Tested and works


src/main/java/frc/robot/commands/ramp/RampMove.java
```diff
@@ -28,6 +28,6 @@ public void execute() {
 
     @Override
     public boolean isFinished() {
-            return true;
+        return (Math.abs(ramp.getRampPos() - encoderValue.getAsDouble()) > 0.05);
```
bengold22: Shouldn't this be <

bengold22 (changes requested):

src/main/java/frc/robot/commands/ramp/RampMove.java
```diff
@@ -28,6 +28,6 @@ public void execute() {
 
     @Override
     public boolean isFinished() {
-            return true;
+        return (Math.abs(ramp.getRampPos() - encoderValue.getAsDouble()) < 0.05);
```
noahheller: make 0,05 a constant

PR #100, opened 2024-02-25
Waiting for motor to wind down in reset ramp

### LINKED ISSUE #96: Fix ResetRamp to wait for .5 seconds after limit switch is reached to reset encoder
Due possibly to behavior of the chain, the "reset ramp" action is not being very consistent. A possible approach to get the encoder to read consistently:

* In isFinished(), mark the time when the limit switch is reached but do not reset the encoder and do not actually finish
* .5 seconds after this point, return true in isFinished

neo-Aditya:
Needs to be tested on the real robot

### DIFF
--- a/src/main/java/frc/robot/commands/ramp/ResetRamp.java
+++ b/src/main/java/frc/robot/commands/ramp/ResetRamp.java
@@ -51,10 +51,14 @@ public class ResetRamp extends Command {
   public boolean isFinished() {
     if (ramp.getReversedSwitchState()) {
       ramp.stopMotor();
-    if(rampStopTime.equals(null)) rampStopTime = Timer.getFPGATimestamp();
-    else if ((Timer.getFPGATimestamp() - rampStopTime) >= Constants.RAMP_WIND_DOWN_TIME) return true;
+      if(rampStopTime.equals(null)) {
+        rampStopTime = Timer.getFPGATimestamp();
+      }
+      else if ((Timer.getFPGATimestamp() - rampStopTime) >= Constants.RAMP_WIND_DOWN_TIME) {
+        return true;
+      }
     }
-    else if ((Timer.getFPGATimestamp() - startTime) >= 5) {
+    else if ((Timer.getFPGATimestamp() - startTime) >= Constants.RESET_RAMP_TIMEOUT) {
       return true;
     }
     return false;
--- a/src/main/java/frc/robot/commands/ramp/ResetRamp.java
+++ b/src/main/java/frc/robot/commands/ramp/ResetRamp.java
@@ -14,6 +14,7 @@ public class ResetRamp extends Command {
   /** Creates a new ResetRamp. */
   private Ramp ramp; 
   private double startTime;
+  private Double rampStopTime = null;
 
   /*
    *When we get the robot:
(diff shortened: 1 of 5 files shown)


src/main/java/frc/robot/commands/ramp/ResetRamp.java
```diff
@@ -41,7 +42,10 @@ public void execute() {}
   @Override
   public void end(boolean interrupted) {
     ramp.stopMotor();
-    ramp.resetEncoder();
+    if(ramp.getReversedSwitchState()) {
+      new WaitCommand(Constants.RAMP_WIND_DOWN_TIME);
```
bengold22: This isn't going to do what you want. new WaitCommand() will create a new wait command, yes, but this method does not take any time. It will immediately return and resetEncoder will run.

bengold22 (changes requested):

src/main/java/frc/robot/constants/Constants2023.java
```diff
@@ -14,6 +14,8 @@ public class Constants2023 extends GameConstants {
     public static final double RAMP_MAX_RPM_VELOCITY = 500;
     public static final double RAMP_MAX_RPM_ACCELERATION = 1500;
     public static final double RESET_RAMP_SPEED = 0.3; //assuming positive is forward, also needs to be refined do the real robot
+    public static final double RAMP_WIND_DOWN_TIME = 0.5; //seconds
```
bengold22: Put this in GameConstants

src/main/java/frc/robot/commands/ramp/ResetRamp.java
```diff
@@ -41,13 +43,21 @@ public void execute() {}
...
+    if(ramp.getReversedSwitchState()) ramp.resetEncoder();
   }
 
   // Returns true when the command should end.
   @Override
   public boolean isFinished() {
-    return ((ramp.getReversedSwitchState()) || ((Timer.getFPGATimestamp() - startTime) >= 5)); //Assuming that Forward Switch is the top one
+    if (ramp.getReversedSwitchState()) {
+      ramp.stopMotor();
+    if(rampStopTime.equals(null)) rampStopTime = Timer.getFPGATimestamp();
```
bengold22: Please format this correctly with braces. We have enough different levels of programmers that this style where the braces are left-off will be highly prone to error.

bengold22 (changes requested):

noahheller:
adding a delay seems like a bandage solution to me and will introduce delay problems for commands that relay on reset ramp to finish. I think we can explore other solutions (like adding a simple PID to slow the motor before resetting to avoid the recoil)

neo-Aditya:
> adding a delay seems like a bandage solution to me and will introduce delay problems for commands that relay on reset ramp to finish. I think we can explore other solutions (like adding a simple PID to slow the motor before resetting to avoid the recoil)

This isn't using PID it's to reset the ramp and the encoder which the PID uses to set position.

noahheller:
I'm aware it's not using the PID but I am suggesting that you use some sort of linear or Pid type de-acceleration to mitigate the bounceback of the ramp affecting encoder reset value without adding a delay

batchen1:
Do we even need this??
I think the drive team should set up the ramp correctly at the beginning of the match. If we need to move it a bit, we can do it at a very low speed until limit switch.
I agree that a 0.5 second delay is not good here, but we can't do a PID either because we don't know where we're going.
Alternatively we can do it in teleopInit() so if drive team didn't reset it properly, it will only affect auto.

PR #107, opened 2024-02-26
Added timeoutCounter and made it so that the intake stops spinning when a piece is seen

neo-Aditya:
Tested and Works


src/main/java/frc/robot/commands/intake/StartIntake.java
```diff
@@ -40,6 +42,6 @@ public boolean isFinished() {
             timeoutCounter.increaseTimeoutCount();
             return true;
         }
-        return false;
+        return(SmartShuffleboard.getBoolean("Driver", "Gamepiece Collected", false));
```
MatthewSpofford, line 45: This is a roundabout way of calling the feeder's `pieceSeen()`. This command should be updated so that the feeder is used directly.

MatthewSpofford (changes requested):

noahheller:
Why do we have this? Don't we always called the intake in a deadline group with the feeder?

src/main/java/frc/robot/commands/intake/StartIntake.java
neo-Aditya: We use the feeder and Intake simultaniously almost all the time so I couldn't find a better way without having to remove the addRequirements in the Feeder commands

neo-Aditya:
> Why do we have this? Don't we always called the intake in a deadline group with the feeder?

I belive it's just to be safe even though we'll almost never use it alone.

src/main/java/frc/robot/commands/intake/StartIntake.java
MatthewSpofford: In this particular case, I don't see anything wrong with the feeder _not_ being a subsystem requirement for this command, since all you need is the color sensor data. So as long as you don't pass the feeder into `addRequirements()`, it should be completely fine to use it without affecting other commands.

noahheller: We need to have a conversation as a team about the purpose of addrequirements and how to manage state. In short, addrequirements only NEEDS to be there if you modify the subsystem, however, I think it is important to discuss the possible consequences if someone tried to modify a subsystem in the command later on. There are many alternatives all with their drawbacks. Some include: passing suppliers into the command, separating logic into other pojos or other subsystems, and some sort of readOnlyProoerty implementation similar to JavaFX using ObservableValues. I don't have the answer but I think we should decide on a constant solution.

PR #78, opened 2024-02-24
Added commands to move arm down and engaging the ratched in AutoInnit

neo-Aditya:
Still not Tested :(

PR #108, opened 2024-02-26
Move arm down plus rachet

### LINKED ISSUE #90: Add a command that puts climbers in the correct state at start of match
* Climbers should be fully retracted
* Servos should be set to "engaged" per hardware request

neo-Aditya:
Here's the og pull request https://github.com/FRC4048/FRC2024_Java/pull/78


noahheller:
This Branch needs to be fixed it containers commits involving other PR that have to do with the ramp

PR #110, opened 2024-02-28
added outtake command (needs to be tested)

noahheller:


src/main/java/frc/robot/commands/feeder/FeederBackEject.java
```diff
@@ -0,0 +1,36 @@
+package frc.robot.commands.feeder;
+
+import edu.wpi.first.wpilibj.Timer;
```
armadilloz, line 3: Is this working? Do we need it?

PR #114, opened 2024-02-28
added commented out functionality to use visualVM (plan on profile wh…

noahheller:
…at is takeing time on robot)


build.gradle
```diff
@@ -27,6 +27,14 @@ deploy {
                 // getTargetTypeClass is a shortcut to get the class type using a string
 
                 frcJava(getArtifactTypeClass('FRCJavaArtifact')) {
+                 // Launch VisualVM with --jdkhome $PATH_TO_WPI_JDK
```
armadilloz, line 30: Have you verified that the profiler even works on the rio Jre?

noahheller: yes it works

armadilloz (approved):

PR #123, opened 2024-03-02
added different commands

Soham-West:


neo-Aditya (dismissed):
Do we even need this? We already have these commands assigned to buttons for operater preferences.

bengold22 (approved):

PR #124, opened 2024-03-02
gave different motor speeds for amp vs shooter

Levercpu:


src/main/java/frc/robot/commands/feeder/FeederGamepieceUntilLeave.java
```diff
@@ -25,7 +28,11 @@ public void end(boolean interrupted) {
 
     @Override
     public void execute() {
-        feeder.setFeederMotorSpeed(Constants.FEEDER_MOTOR_EXIT_SPEED);
+        if (ramp.getRampPos()<Constants.RAMP_POS_SHOOT_AMP) {
+            feeder.setFeederMotorSpeed(Constants.FEEDER_MOTOR_EXIT_SPEED);
+        } else {
+            feeder.setFeederMotorSpeed(Constants.FEEDER_MOTOR_EXIT_SPEED*0.7);
+        }
```
bengold22, line 35: This needs to be another constant, this only works when the original value was 1.0 (which it might not be in the future).

bengold22:
Many of these sequences you had to change have now been deleted. You'll want to merge from main and don't re-add the files that were deleted, because they have been intentionally removed.

bengold22 (changes requested):

armadilloz (approved):

PR #126, opened 2024-03-02
title self explain

Levercpu:


src/main/java/frc/robot/commands/climber/ManualControlClimber.java
```diff
@@ -24,7 +24,7 @@ public void end(boolean interrupted) {
     @Override
     public void execute() {
         double value = supplier.getAsDouble();
-        if (Math.abs(value) > .2) {
+        if (Math.abs(value) > .2 && climber.getEncoderPosition() < 80 && climber.isExtendedLimitSwitchNotPressed()) {
```
batchen1: 1 - no need to check the limit switch because it is connected directly to the motor controller.
2 - this is probably not the best place to put this code, there are 2 encoders, one for each arm, and this code will not work. Can you move the logic to the subsystem setSpeed method?

batchen1 (changes requested):

src/main/java/frc/robot/subsystems/Climber.java
```diff
@@ -55,7 +55,7 @@ public Climber() {
      * @return true if setting speed was successful
      */
     public boolean setSpeed(double spd) {
-        if (spd > 0 && ratchetEngaged) return false;
+        if (spd > 0 && (ratchetEngaged || climberRight.getEncoder().getPosition() > 80 || climberLeft.getEncoder().getPosition() > 80)) return false;
```
batchen1: this is still wrong, if any one of the encoders is above 80 none of them will move.

batchen1 (changes requested):

src/main/java/frc/robot/subsystems/Climber.java
bengold22: This code needs to be more like this:

```
if (spd >0 && ratchEngaged) return false; // This is needed always, cannot raise when ratchets are engaged

// Now we check motor limits
if (spd > 0) {
  if (climberRight.getEncoder().getPosition() > 80) {
    climberRight.set(0);
  } else {
    climberRight.set(-spd);
  }
  if (climberLeft.getEncoder().getPosition() > 80) {
    climberLeft.set(0);
  } else {
    climberLeft.set(spd);
  }
} else {
  // Down can always run to limit switches, these aren't as sensitive
```

bengold22 (changes requested):

src/main/java/frc/robot/subsystems/Climber.java
bengold22: Also the right climber (because it's moving negatively), will need the check to be:

if (climberRight.getEncoder().getPosition() < -80) {

src/main/java/frc/robot/subsystems/Climber.java
```diff
@@ -77,6 +89,12 @@ public void disengageRatchet() {
         rightServo.setPosition(0);
         ratchetEngaged = false;
     }
+    public double getEncoderPosition() {
+        return climberLeft.getEncoder().getPosition();
+    }
+    public boolean isExtendedLimitSwitchNotPressed() {
```
neo-Aditya: Can you add a isExtendedLimitSwitchPressed instead aswell as one for Left and Right seperately, though the latter my not be needed nor wanted.

neo-Aditya (changes requested):

armadilloz (approved):

PR #128, opened 2024-03-03
Delted Unused Commands and Unneeded timeouts

Sahiltheram:


src/main/java/frc/robot/commands/ramp/ManualRamp.java
batchen1, line 1: Let's not delete this one? it is useful if we want to adjust the shooting angle.

batchen1 (changes requested):

batchen1 (approved):

PR #135, opened 2024-03-03
Fixed Turned Gamepiece

Sahiltheram:


neo-Aditya:
Wow Sam you did so much!

src/main/java/frc/robot/constants/GameConstants.java
```diff
@@ -9,15 +9,15 @@ public class GameConstants {
     public static final double AUTO_SPOOL_AND_SHOOT_TIME = 4;
     //DEBUGS
     public static final boolean SHOOTER_DEBUG = false;
-    public static final boolean RAMP_DEBUG = false;
-    public static final boolean RAMP_PID_DEBUG = false;
+    public static final boolean RAMP_DEBUG = true;
```
neo-Aditya: Set all these debugs to false

src/main/java/frc/robot/commands/MoveToGamepiece.java
```diff
@@ -13,48 +13,46 @@ public class MoveToGamepiece extends Command {
...
-        else driveStates = drivetrain.createChassisSpeeds(-0.6, 0.0, 0.0, false);
+        driveStates = new ChassisSpeeds(movingPIDController.calculate(ychange), 0, turningPIDController.calculate(vision.getPieceOffestAngleX() - Constants.LIMELIGHT_TURN_TO_PIECE_DESIRED_X));
         drivetrain.drive(driveStates);
+        }
     }
 
     @Override
     public boolean isFinished() {
-        return ((Timer.getFPGATimestamp() - timeSincePieceLoss >= Constants.TIMEOUT_AFTER_PIECE_NOT_SEEN) || (Timer.getFPGATimestamp() - startTime > Constants.MOVE_TO_GAMEPIECE_TIMEOUT));
+        return (Timer.getFPGATimestamp() - startTime > Constants.MOVE_TO_GAMEPIECE_TIMEOUT) || (!vision.isPieceSeen() || (Math.abs(ychange) < Constants.MOVE_TO_GAMEPIECE_THRESHOLD));
```
neo-Aditya: Can you add a Timeout counter to this so that we can log when it times out.

src/main/java/frc/robot/commands/MoveToGamepiece.java
```diff
@@ -13,48 +13,46 @@ public class MoveToGamepiece extends Command {
...
+    private double ychange;
+
 
     public MoveToGamepiece(SwerveDrivetrain drivetrain, Vision vision) {
         this.drivetrain = drivetrain;
         this.vision = vision;
         addRequirements(drivetrain);
         TrapezoidProfile.Constraints constraints = new TrapezoidProfile.Constraints(Constants.GAMEPIECE_MAX_VELOCITY, Constants.GAMEPIECE_MAX_ACCELERATION);
-        double tP = 0.04;
+        double tP = 0.015;
```
neo-Aditya: Can you make these constants

neo-Aditya (dismissed):

src/main/java/frc/robot/constants/GameConstants.java
batchen1: false

src/main/java/frc/robot/constants/GameConstants.java
```diff
@@ -9,15 +9,15 @@ public class GameConstants {
...
+    public static final boolean RAMP_PID_DEBUG = true;
```
batchen1: same

src/main/java/frc/robot/constants/GameConstants.java
```diff
@@ -9,15 +9,15 @@ public class GameConstants {
...
     public static final boolean SWERVE_DEBUG = false;
     public static final boolean FEEDER_DEBUG = false;
     public static final boolean CLIMBER_DEBUG = false;
     public static final boolean INTAKE_DEBUG = false;
     public static final boolean DEPLOYER_DEBUG = false;
     public static final boolean AMP_DEBUG = false;
-    public static final boolean VISION_DEBUG = false;
+    public static final boolean VISION_DEBUG = true;
```
batchen1: same

src/main/java/frc/robot/commands/MoveToGamepiece.java
batchen1: don't add a timeoutConter here, the goal of the timeoutCounter is to detect a bad sensor. This is not the case here, this command may time out and it is ok.

src/main/java/frc/robot/commands/MoveToGamepiece.java
```diff
@@ -13,48 +13,46 @@ public class MoveToGamepiece extends Command {
...
     public void execute() {
-        ChassisSpeeds driveStates;
-        double ychange = vision.getPieceOffestAngleY() - Constants.LIMELIGHT_TURN_TO_PIECE_DESIRED_Y;
+        ychange = vision.getPieceOffestAngleY() - Constants.LIMELIGHT_TURN_TO_PIECE_DESIRED_Y;
         if (vision.isPieceSeen() && (Math.abs(ychange) > Constants.MOVE_TO_GAMEPIECE_THRESHOLD)) {
-            timeSincePieceLoss = Timer.getFPGATimestamp();
-            driveStates = new ChassisSpeeds(movingPIDController.calculate(ychange), 0, turningPIDController.calculate(vision.getPieceOffestAngleX()));
-        }
-        else driveStates = drivetrain.createChassisSpeeds(-0.6, 0.0, 0.0, false);
+        driveStates = new ChassisSpeeds(movingPIDController.calculate(ychange), 0, turningPIDController.calculate(vision.getPieceOffestAngleX() - Constants.LIMELIGHT_TURN_TO_PIECE_DESIRED_X));
```
batchen1, line 44: We don't want to end the command if we don't see a game piece momentarily. Can you add back the counter that counts how many consecutive cycles with no piece, and end the command only if we reach some threshold.

batchen1 (changes requested):

src/main/java/frc/robot/commands/MoveToGamepiece.java
Sahiltheram: Done

batchen1: you need to reset the counter if you see a game piece.

src/main/java/frc/robot/commands/MoveToGamepiece.java
```diff
@@ -8,53 +8,63 @@
...
+        if (!vision.isPieceSeen()) {
+            ++cycle;
+        }
     }
 
     @Override
     public boolean isFinished() {
-        return ((Timer.getFPGATimestamp() - timeSincePieceLoss >= Constants.TIMEOUT_AFTER_PIECE_NOT_SEEN) || (Timer.getFPGATimestamp() - startTime > Constants.MOVE_TO_GAMEPIECE_TIMEOUT));
+        if ((Timer.getFPGATimestamp() - startTime > Constants.MOVE_TO_GAMEPIECE_TIMEOUT) || (Math.abs(ychange) < Constants.MOVE_TO_GAMEPIECE_THRESHOLD)) {
+            timeoutCounter.increaseTimeoutCount();
```
batchen1: Please delete the timoutCounter. We don't want to increase it here.

batchen1 (changes requested):

src/main/java/frc/robot/commands/MoveToGamepiece.java
batchen1: This is still incorrect. You're always resetting the counter so it will never be >5.
Should be something like:
if (pieceSeen) { cycle = 0;} else (cycle++;)

Sahiltheram: Hopefully it is good now

batchen1 (approved):

src/main/java/frc/robot/commands/MoveToGamepiece.java
neo-Aditya: We don't want a variable to count the number of cycles seen since the Vision subsystem does that for you

bengold22 (approved):

PR #145, opened 2024-03-05
Autochooser nameing

noahheller:
also has changes to stop and unregister commands when disabled and in test mode.


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -305,4 +301,24 @@ public static boolean isRedAlliance() {
     public AutoChooser getAutoChooser() {
         return autoChooser;
     }
+
+    public void regesterDefaultCommands() {
+        drivetrain.setDefaultCommand(new Drive(drivetrain, joyleft::getY, joyleft::getX, joyright::getX));
+        climber.setDefaultCommand(new ManualControlClimber(climber, () -> -controller.getLeftY()));
+    }
+    public void unregesterDefaultCommands() {
```
MatthewSpofford, line 309: It doesn't look like this method is used anywhere. Is that intended?

batchen1 (dismissed):

PR #147, opened 2024-03-05
Nh wpi fixes

noahheller:


batchen1 (dismissed):

src/main/java/frc/robot/utils/AutoAlignment.java
```diff
@@ -61,19 +80,33 @@ public static Rotation2d getAngle(Alignable alignable, double x, double y) {
...
+    public static Rotation2d getYaw(Alignable alignable, double x, double y, double z, double rampXOffset) {
         if (isInvalidAngle(alignable)) return new Rotation2d();
         BiFunction<Double, Double, Rotation2d> function = positionYawMap.get(alignable);
         if (isInvalidFunction(function)) return new Rotation2d();
-        double xNorm = x + (RobotContainer.isRedAlliance() ? -Constants.RAMP_X_OFFSET : Constants.RAMP_X_OFFSET);
+        double xNorm = x + (RobotContainer.isRedAlliance() ? rampXOffset: -rampXOffset);
         double deltaX = xNorm - alignable.getX();
         double deltaY = y - alignable.getY();
         double dist = Math.hypot(deltaX, deltaY);
+        SmartDashboard.putNumber("DISTANCE_XY", dist);
```
armadilloz: Remove (or put in debug)

src/main/java/frc/robot/subsystems/SwerveDrivetrain.java
```diff
@@ -69,12 +69,12 @@ public void periodic() {
 
 
         gyroValue = getGyro();
-        if (SmartDashboard.getBoolean("USE VISION",false)){
-            poseEstimator.updatePositionWithVis(gyroValue);
-        }else {
-            poseEstimator.updatePosition(gyroValue);
-        }
-
+//        if (SmartDashboard.getBoolean("USE VISION",true)){
```
armadilloz: I don't understand this: Shouldn't we always update with Gyro?
Shouldn't we leave the boolean to control updates from vision?

src/main/java/frc/robot/constants/GameConstants.java
```diff
@@ -145,8 +145,8 @@ public class GameConstants {
     public static final int TIMED_INTAKE_AUTO_TIMEOUT = 2;
 
     public static final double HIGHT_OF_RAMP = 0.66; // needs to be measured
-    public static final double RAMP_MIN_ANGLE = 32.5; //degrees
-    public static final double RAMP_MAX_ANGLE = 90; //degrees
-    public static final double RAMP_X_OFFSET = 0.3; //meters, tip of ramp offset from center of robot
+    public static final double RAMP_MIN_ANGLE = 31; //degrees
+    public static final double RAMP_MAX_ANGLE = 52; //degrees
+//    public static final double RAMP_X_OFFSET = 0.43 - 0.3; //meters, tip of ramp offset from center of robot (take robot radians and subtract dist from edge (with bumpers)
```
armadilloz: Remove?

src/main/java/frc/robot/commands/shooter/AdvancedSpinningShot.java
```diff
@@ -45,7 +45,7 @@ public void end(boolean interrupted) {
 
     @Override
     public boolean isFinished() {
-        return alignable == null || alignableSupplier.get() == null || !alignable.equals(alignableSupplier.get());
+        return false;
```
armadilloz: why remove the termination condition?

src/main/java/frc/robot/commands/ramp/RampFollow.java
```diff
@@ -47,6 +50,8 @@ public void execute() {
 
     @Override
     public boolean isFinished() {
-        return alignable == null || alignableSupplier.get() == null || !alignable.equals(alignableSupplier.get());
+        boolean isFinished = alignable == null || alignableSupplier.get() == null || !alignable.equals(alignableSupplier.get());
+        SmartDashboard.putBoolean("RampDone", isFinished);
```
armadilloz: if debug?

armadilloz: Also - this would appear in logs no no real need

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -277,15 +286,13 @@ private void configureBindings() {
...
-                        new RampMove(ramp, () -> GameConstants.RAMP_POS_STOW)
-                ),
-                new AdvancedSpinningShot(shooter,() -> drivetrain.getPose(), ()-> drivetrain.getAlignable())
-        ));
+        SequentialCommandGroup advancedShoot = new SequentialCommandGroup(
+                new FeederGamepieceUntilLeave(feeder, ramp),
+                new WaitCommand(GameConstants.SHOOTER_TIME_BEFORE_STOPPING),
+                new RampMove(ramp, () -> GameConstants.RAMP_POS_STOW)
+        );
+        controller.povRight().onTrue(advancedShoot);
```
armadilloz: I don't think we should use the same command instance in two bindings - there can be state changes that are unpredictable from the command being scheduled from two different places.

armadilloz (changes requested):

src/main/java/frc/robot/subsystems/SwerveDrivetrain.java
noahheller: The original purpose of the USE VISION was to disable it until we were done testing. Because vision was working I got rid of the USE VISION. If we want to add it back however I think it would be fine (probably want to make it a constant instead of a shuffleboard value)

src/main/java/frc/robot/commands/shooter/AdvancedSpinningShot.java
noahheller: How it was implemented when we went to WPI was that the advanced shot was independent of the alingable's supplier current value, so they did not want the command to end when they released the trigger to align to the speaker. However now that the advanced shot is tied to current alignable being held I could add the condition back.

src/main/java/frc/robot/commands/ramp/RampFollow.java
noahheller: removed

src/main/java/frc/robot/RobotContainer.java
noahheller: removed from operator controller

src/main/java/frc/robot/constants/GameConstants.java
noahheller: done

src/main/java/frc/robot/utils/AutoAlignment.java
noahheller: done

src/main/java/frc/robot/subsystems/SwerveDrivetrain.java
armadilloz: Constant is fine. We should keep a feature flag for it somehow

armadilloz (approved):

PR #152, opened 2024-03-09
Nh wb log command

bengold22:


noahheller:
Do we still want this?

PR #153, opened 2024-03-11
multithreading color sensor

noahheller:
Needs to be tested.

### DIFF
--- a/src/main/java/frc/robot/subsystems/Feeder.java
+++ b/src/main/java/frc/robot/subsystems/Feeder.java
@@ -34,7 +34,9 @@ public class Feeder extends SubsystemBase {
         Robot.getDiagnostics().addDiagnosable(new DiagColorSensor("Feeder", "Color Sensor", colorSensor));
         new Thread(() -> {
             try {
-                colorReadingBuffer.put(colorSensor.getMatchedColor());
+                while (true) {
+colorReadingBuffer.put(colorSensor.getMatchedColor());
+                }
             } catch (InterruptedException e) {
                 DriverStation.reportError("Color Sensor put call interrupted", e.getStackTrace());
             }
--- a/src/main/java/frc/robot/subsystems/Feeder.java
+++ b/src/main/java/frc/robot/subsystems/Feeder.java
@@ -71,6 +71,7 @@ public class Feeder extends SubsystemBase {
     @Override
     public void periodic() {
         thisTicMaxConfidence = maxConfidence.get();
+        maxConfidence.set(0);
         if (Constants.FEEDER_DEBUG) {
             ColorValue detectedColor = colorSensor.getColor();
             Color rawColor = colorSensor.getRawColor();
--- a/src/main/java/frc/robot/subsystems/Feeder.java
+++ b/src/main/java/frc/robot/subsystems/Feeder.java
@@ -3,7 +3,6 @@ package frc.robot.subsystems;
 import com.ctre.phoenix.motorcontrol.NeutralMode;
 import com.ctre.phoenix.motorcontrol.can.WPI_TalonSRX;
 import com.revrobotics.ColorMatchResult;
-import edu.wpi.first.wpilibj.DriverStation;
 import edu.wpi.first.wpilibj.I2C;
 import edu.wpi.first.wpilibj.util.Color;
 import edu.wpi.first.wpilibj2.command.SubsystemBase;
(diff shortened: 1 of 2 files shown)


goatfanboi23:
TODO: use something other than while true. (Don't want to code on phone right now)

src/main/java/frc/robot/subsystems/Feeder.java
```diff
@@ -13,18 +14,33 @@
...
 
     public Feeder() {
         this.feederMotor = new WPI_TalonSRX(Constants.FEEDER_MOTOR_ID);
         this.feederMotor.setNeutralMode(NeutralMode.Brake);
-
         colorSensor = new ColorSensor(i2cPort);
         Robot.getDiagnostics().addDiagnosable(new DiagColorSensor("Feeder", "Color Sensor", colorSensor));
+        new Thread(() -> {
+            try {
+                while (true) {
```
armadilloz: A tight loop like that would kill the CPU. Need to insert a sleep to reduce the frequency of polling to the desired one.

src/main/java/frc/robot/subsystems/Feeder.java
```diff
@@ -13,18 +14,33 @@
...
+import java.util.ArrayList;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+
 public class Feeder extends SubsystemBase {
 
     private final WPI_TalonSRX feederMotor;
     private final I2C.Port i2cPort = I2C.Port.kMXP;
     private final ColorSensor colorSensor;
+    private final BlockingQueue<ColorMatchResult> colorReadingBuffer = new LinkedBlockingQueue<>();
```
armadilloz: Unbound blocking queue would eat up too much memory in case we fail to drain it quickly enough

armadilloz: Blocking queue is not right here (if bound in size) since it will block the writing if full and the reading if empty.

armadilloz: An alternative solution that does not involve a queue is to use an AtomicInteger that represents the Maximum certainty or an AtomicBoolean that represents a find. This way you don't consume memory and do not block.

Another alternative is to store just the raw reading rather than the matched one, since the matching can take a lot of compute, so we can do the matching only when needed.

src/main/java/frc/robot/subsystems/Feeder.java
```diff
@@ -47,13 +63,19 @@ public ColorValue getColor() {
...
+                return true;
+            }
+        }
+        return false;
     }
 
     @Override
     public void periodic() {
+        resultsSinceLastTic.clear();
+        colorReadingBuffer.drainTo(resultsSinceLastTic);
```
armadilloz: This assumes that the periodic and pieceSeen run in the same thread (otherwise you'll get a concurrentModificationException) which may be correct assumption but still...

armadilloz (changes requested):
Overall cool approach! Nice job. 
The issue I have is that we don't really know what the problem with the intake is (at least, I haven't seen anything that analyzes it). This solution assumes that the detection is too late when the periodic actually runs, which may be part of the problem but it won't solve the case that there was a detection but the piece keeps on going after the motor stop call because of momentum. I think more research on that would be appropriate.
We can certainly experiment with this solution though to see if it is better.

src/main/java/frc/robot/subsystems/Feeder.java
noahheller: I like the idea of an AtomicIntereger.  I feel like storing the raw reading and then having to match the color in the main thread would slow things down, although that is just a hunch. My original idea was to use an atomic boolean however I did not like the idea of having an AtomicBoolean because we have different confidence thresholds for incoming and exiting feeder.

noahheller:
> Overall cool approach! Nice job. The issue I have is that we don't really know what the problem with the intake is (at least, I haven't seen anything that analyzes it). This solution assumes that the detection is too late when the periodic actually runs, which may be part of the problem but it won't solve the case that there was a detection but the piece keeps on going after the motor stop call because of momentum. I think more research on that would be appropriate. We can certainly experiment with this solution though to see if it is better.

The reason I believe the color sensor is noting being updated fast enough is because the startFeeder command is timeing out durring auto meaning it was not tripped when the gamepiece was intaked. In terms of the momentum problem hopefully we can fix that with modifying motor speed.

src/main/java/frc/robot/subsystems/Feeder.java
noahheller: done

src/main/java/frc/robot/subsystems/Feeder.java
noahheller: replaced with atomic integer

noahheller: replaced with atomic integer

src/main/java/frc/robot/subsystems/Feeder.java
noahheller: I set variable to an atomic integer. 
One thing to keep in mind if we run into problems is that there is a period between the subsystem periodic (called first) and any commands isFinished calls that read the color sensor. So if let's say the piece enters the sensor but feeder periodic has already said that the piece is not there, then the motors would not stop for another tic. I don't think this will be noticeable (1/50 of a second delay should be fine) but I thought I would bring it up.

src/main/java/frc/robot/subsystems/Feeder.java
```diff
@@ -47,13 +64,13 @@ public ColorValue getColor() {
...
     public boolean pieceSeen(boolean incoming) {
-        ColorMatchResult latestResult = colorSensor.getMatchedColor();
         double confidence = incoming ? Constants.COLOR_CONFIDENCE_RATE_INCOMING : Constants.COLOR_CONFIDENCE_RATE_BACKDRIVE;
-        return latestResult.confidence > confidence;
+        return thisTicMaxConfidence > (confidence * 100);
     }
 
     @Override
     public void periodic() {
+        thisTicMaxConfidence = maxConfidence.get();
```
armadilloz, line 73: Need to reset the max confidence to 0

src/main/java/frc/robot/subsystems/Feeder.java
armadilloz: The Eason for delaying the color matching is to do it once per cycle rather than every time we read the value.
We can do that later

src/main/java/frc/robot/subsystems/Feeder.java
noahheller: done

PR #154, opened 2024-03-12
Nh power

noahheller:


src/main/java/frc/robot/subsystems/PowerMonitor.java
```diff
@@ -0,0 +1,155 @@
...
+        this.pdh = new PowerDistribution(1, PowerDistribution.ModuleType.kRev);
+    }
+
+    /**
+     * @param channel to monitor
+     * @return boolean representing if addition to monitoring set was successful
+     */
+    public boolean logChannel(int channel) {
+        if (channel < pdh.getNumChannels()) {
+            return monitoringChannels.add(channel);
```
armadilloz: A `BitSet` may be more appropriate

src/main/java/frc/robot/subsystems/PowerMonitor.java
```diff
@@ -0,0 +1,155 @@
...
+    /**
+     * Runs action when channel current draw reaches a threshold. <br>
+     * Note, when channel reaches current draw you will have to re-register the event
+     *
+     * @param channel          to monitor
+     * @param currentThreshold current threshold before runnable is called
+     * @param runnable         action to preform when current is reached
+     * @return if addition to currentEvent list was successful
+     */
+    public boolean onCurrentAboveThreshold(int channel, double currentThreshold, Runnable runnable) {
```
armadilloz: `above` and `below` can be replaced with a predicate e.g. `Predicate<current>` to let the caller decide whether they want to trigger or not.

src/main/java/frc/robot/subsystems/PowerMonitor.java
```diff
@@ -0,0 +1,155 @@
...
+        return pdh.getVoltage();
+    }
+
+    private void callCurrentMaxEvents() {
+        List<ChannelCurrentEvent> events = maxCurrentEvents.stream().filter(event -> {
+            int channel = event.getChannel();
+            return getChannelCurrent(channel) > event.getCurrentThreshold();
+        }).toList();
+        events.forEach(e -> e.getAction().run());
+        maxCurrentEvents.removeAll(events);
```
armadilloz: Instead of a `Runnable` you can get a BiFunction<channel, current, Boolean> that will return `true` if it cares to stay registered. This will save the need to re-register every time.

src/main/java/frc/robot/subsystems/PowerMonitor.java
armadilloz: Probably need a way to unsubscribe too.

armadilloz (dismissed):

src/main/java/frc/robot/subsystems/PowerMonitor.java
noahheller: done

src/main/java/frc/robot/subsystems/PowerMonitor.java
noahheller: done

src/main/java/frc/robot/subsystems/PowerMonitor.java
noahheller: done

src/main/java/frc/robot/subsystems/PowerMonitor.java
noahheller: done

bengold22 (approved):

PR #176, opened 2024-03-21
Added diag for the beam, going for a singular boolean version

neo-Aditya:


noahheller:
This will trip if two of the beams trip (due to custom circut). Is this the desired result? we will not know if a beam is broken?

batchen1 (approved):

PR #178, opened 2024-03-22
Nh command loop

noahheller:
contains a couple of optimizations that might increase performance. If we dont like any of them they are separated by commits, so we can cherrypick


src/main/java/frc/robot/subsystems/Deployer.java
```diff
@@ -70,13 +70,11 @@ public void periodic() {
...
+                    .withSize(2, 2);
         }
-        
-        /*
-        SmartShuffleboard.put("Driver", "Is Intake Deployer Raised?", isDeployerForwardLimitSwitchClosed())
-            .withPosition(0, 2)
-            .withSize(2, 2);
-        */
+
         Logger.logBoolean(baseLogName + "FWD LMT", isDeployerForwardLimitSwitchClosed(), Constants.ENABLE_LOGGING);
```
armadilloz: Move to the if (debug)?

src/main/java/frc/robot/subsystems/Feeder.java
```diff
@@ -77,13 +77,11 @@ public void periodic() {
...
+                    .withPosition(0, 0)
+                    .withSize(2, 2);
         }
 
-        /*
-        SmartShuffleboard.put("Driver", "Has Game Piece?", pieceSeen(false))
-            .withPosition(0, 0)
-            .withSize(2, 2);
-        */
         Logger.logDouble(baseLogName + "FeederMotorSpeed",getFeederMotorSpeed(),Constants.ENABLE_LOGGING);
```
armadilloz: Move to if(debug)?

src/main/java/frc/robot/subsystems/Ramp.java
```diff
@@ -38,35 +38,23 @@ public void periodic() {
             SmartShuffleboard.put("Ramp", "Desired pos", rampPos);
             SmartShuffleboard.put("Ramp", "Reverse Switch Tripped", getReversedSwitchState());
             SmartShuffleboard.put("Ramp", "Forward Switch Tripped", getForwardSwitchState());
-//            double pidP = SmartShuffleboard.getDouble("Ramp", "PID P", neoPidMotor.getPidController().getP());
```
armadilloz, line 49: The code above (setting thePID_FF) can probably be be part of a command rather then run every time in periodic

src/main/java/frc/robot/subsystems/Ramp.java
```diff
@@ -38,35 +38,23 @@ public void periodic() {
...
-//            double pidI = SmartShuffleboard.getDouble("Ramp", "PID I", neoPidMotor.getPidController().getI());
-//            double pidD = SmartShuffleboard.getDouble("Ramp", "PID D", neoPidMotor.getPidController().getD());
-//            double pidFF = SmartShuffleboard.getDouble("Ramp", "PID FF", neoPidMotor.getPidController().getFF());
-//            if (pidP != neoPidMotor.getPidController().getP()) neoPidMotor.getPidController().setP(pidP);
-//            if (pidI != neoPidMotor.getPidController().getI()) neoPidMotor.getPidController().setI(pidI);
-//            if (pidD != neoPidMotor.getPidController().getD()) neoPidMotor.getPidController().setD(pidD);
-//            if (pidFF != neoPidMotor.getPidController().getFF()) neoPidMotor.getPidController().setFF(pidFF);
-            SmartShuffleboard.put("Ramp", "Forward Switch Tripped", getForwardSwitchState());
```
armadilloz, line 57: why deleting the debug shuffleboard values?

src/main/java/frc/robot/subsystems/Ramp.java
```diff
@@ -38,35 +38,23 @@ public void periodic() {
...
+                    .withSize(1, 1);
+            SmartShuffleboard.put("Driver", "Speaker Away", isShootAwayAngle())
+                    .withPosition(9, 1)
+                    .withSize(1, 1);
+            SmartShuffleboard.put("Driver", "Amp", isShootAmpAngle())
+                    .withPosition(8, 1)
+                    .withSize(1, 1);
         }
 
         Logger.logDouble(baseLogName + "EncoderValue", getRampPos(), Constants.ENABLE_LOGGING);
```
armadilloz: Move to if(debug)?

src/main/java/frc/robot/subsystems/SwerveDrivetrain.java
```diff
@@ -96,6 +96,7 @@ public void periodic() {
             SmartShuffleboard.put("DriveTrain", "Total Drive Current", totalDriveCurrent);
             SmartShuffleboard.put("DriveTrain", "Total Steer Current", totalSteerCurrent);
             SmartShuffleboard.put("DriveTrain", "TOTAL Current", totalCurrent);
+            SmartShuffleboard.put("GYRO", "Gyro Angle", gyroValue);
```
armadilloz: Logging the Byron before calculating?

src/main/java/frc/robot/subsystems/SwerveDrivetrain.java
```diff
@@ -105,9 +106,6 @@ public void periodic() {
         }else {
             poseEstimator.updatePosition(gyroValue);
         }
-        if (Constants.SWERVE_DEBUG) {
```
armadilloz: lines 103-107 seem expensive, maybe put some in separate thread?

src/main/java/frc/robot/subsystems/LightStrip.java
```diff
@@ -1,49 +1,39 @@
...
-                }
-        );
+                            setPattern(predicateLightEvents.get(c));
+                            predicateLightEvents.remove(c);
+                        }
+                ),0,40, TimeUnit.MILLISECONDS);
     }
 
-    public void setPattern(BlinkinPattern pattern) {
+    public synchronized void setPattern(BlinkinPattern pattern) {
```
armadilloz: Avoid `synchronized` in robot code. There should be no reason for it.

src/main/java/frc/robot/subsystems/LightStrip.java
```diff
@@ -1,49 +1,39 @@
...
-
     public LightStrip(int port) {
         this.colorSensorPort = new Spark(port);
-    }
-
-    @Override
-    public void periodic() {
-        predicateLightEvents.keySet()
+        Executors.newScheduledThreadPool(1).scheduleAtFixedRate(() -> predicateLightEvents.keySet()
                 .stream()
```
armadilloz: This is likely wasteful in new classes and garbage collection. Also, instead of t a new executor we may rely on a simpler mechanism

src/main/java/frc/robot/subsystems/LightStrip.java
```diff
@@ -1,49 +1,39 @@
...
+                            predicateLightEvents.remove(c);
+                        }
+                ),0,40, TimeUnit.MILLISECONDS);
     }
 
-    public void setPattern(BlinkinPattern pattern) {
+    public synchronized void setPattern(BlinkinPattern pattern) {
         this.pattern = pattern;
         colorSensorPort.set(pattern.getPwm());
-        SmartDashboard.putString("PATTERN",pattern.toString());
```
armadilloz, line 41: Put in if(debug) instead of removing?

src/main/java/frc/robot/commands/MoveToGamepiece.java
```diff
@@ -9,44 +9,46 @@
...
     @Override
     public void execute() {
-        ychange = vision.getPieceOffestAngleY() - Constants.LIMELIGHT_MOVE_TO_PIECE_DESIRED_Y;
-        xchange = vision.getPieceOffestAngleX() - Constants.LIMELIGHT_MOVE_TO_PIECE_DESIRED_X;
-        if (vision.isPieceSeen() && (ychange > Constants.MOVE_TO_GAMEPIECE_THRESHOLD)) {
-        driveStates = new ChassisSpeeds(movingPIDController.calculate(ychange), 0, turningPIDController.calculate(xchange));
-        drivetrain.drive(driveStates);
+        yChange = vision.getPieceOffestAngleY() - Constants.LIMELIGHT_MOVE_TO_PIECE_DESIRED_Y;
+        double xChange = vision.getPieceOffestAngleX() - Constants.LIMELIGHT_MOVE_TO_PIECE_DESIRED_X;
+        if (yChange == 0){
```
armadilloz: That does not look right - that should use the `tv`
I think the logic for `isPieceSeen` should have been copied over.

armadilloz (changes requested):

src/main/java/frc/robot/subsystems/LightStrip.java
noahheller: done

src/main/java/frc/robot/subsystems/SwerveDrivetrain.java
noahheller: putting in another branch

src/main/java/frc/robot/subsystems/Ramp.java
noahheller: delete selection was to big when removing pid stuff

noahheller: fixed

src/main/java/frc/robot/subsystems/LightStrip.java
noahheller: done

src/main/java/frc/robot/subsystems/SwerveDrivetrain.java
noahheller: moved to top

src/main/java/frc/robot/subsystems/Ramp.java
noahheller: done

src/main/java/frc/robot/subsystems/Feeder.java
noahheller: done

src/main/java/frc/robot/subsystems/Deployer.java
noahheller: I feel like the benefit of leaving it (what if the switch gets bent and the robot is not hitting the switch?) is greater than the performance gain.

src/main/java/frc/robot/subsystems/LightStrip.java
noahheller: made into a thread

src/main/java/frc/robot/subsystems/Ramp.java
noahheller: Since the ramp is a PID and the MoveRamp command finishes instantly, what do you recommend? Reading the document it does say to use the method "Infrequently" because it takes time. I could store the last PID Value and compare against it before setting it?

src/main/java/frc/robot/commands/MoveToGamepiece.java
noahheller: I think it is better now

src/main/java/frc/robot/subsystems/SwerveDrivetrain.java
noahheller: (the gyro is in another branch) the pose estimator would be hard to put into another thread because wpilib is not know to be thread safe and calls a lot of built in methods. I'm sure its possible but hopefully the pose estimator is not taking to much time. However, I can take a brief look if there is a section of it that would be easy to thread

src/main/java/frc/robot/subsystems/LightStrip.java
```diff
@@ -2,41 +2,58 @@
...
+                    throw new RuntimeException(e);
+                }
+            }
+        }).start();
     }
 
-    public synchronized void setPattern(BlinkinPattern pattern) {
-        this.pattern = pattern;
+    public void setPattern(BlinkinPattern pattern) {
+        this.pattern.set(pattern);
```
armadilloz, line 40: Seems like there is no real need to protect the `pattern` and the `running` since they are primitives (boolean and a reference), though using an atomic reference here does not hurt. The one thing that does require protecting as it is modified from both threads is the map, which is currently vulnerable.

src/main/java/frc/robot/subsystems/Vision.java
```diff
@@ -53,6 +53,9 @@ public double getPieceOffestAngleX() {
   public double getPieceOffestAngleY() {
     return ty.getDouble(0);
   }
+  public double getTv(){
```
armadilloz: A better name would be `isPieceSeen`. This one sounds like it porvides you with an appliance.

src/main/java/frc/robot/subsystems/SwerveDrivetrain.java
```diff
@@ -69,6 +69,7 @@ private double getGyro() {
 
     @Override
     public void periodic() {
+        gyroValue = getGyro();
         frontLeftDriveCurrent = ((CANSparkMax)frontLeft.getSwerveMotor().getDriveMotor()).getOutputCurrent();
```
armadilloz: All the Current stuff can be placed inside `if(CURRENT_DEBUG)`. It can be set to `true` is we want it in the competition, but this way we at least have some control over it

src/main/java/frc/robot/subsystems/Ramp.java
armadilloz: Still some logs left outside the condition

src/main/java/frc/robot/subsystems/Deployer.java
armadilloz: I don't think anyone is looking at these values during the competition (we have diags for that), and even if they are important, they can be placed behind another LOGGING constant that can default to TRUE, so we can control it.

src/main/java/frc/robot/subsystems/Ramp.java
armadilloz: Is there a functional reason that the command be done immediately? I would think that it would make sense for a command to terminate once the ramp it at the position.
So you can have a sequential group of two commands: One to get the ramp "close enough" with FF and another to get the ramp closer (without the FF), and the second one to end once the ramp error is within range.

armadilloz (changes requested):
Mostly fixed, a few more comments.
Plus, I don't see anything dangerous here that should be kept away, with the potential exception of the Limelight piece command that seems to have functional changes.

src/main/java/frc/robot/subsystems/Deployer.java
noahheller: done

src/main/java/frc/robot/subsystems/Ramp.java
noahheller: done

src/main/java/frc/robot/subsystems/Ramp.java
noahheller: I checked and I don't _think_ it will break anything if I make the ramp move not end immediately.  I added the logic to the ramp commands

src/main/java/frc/robot/subsystems/LightStrip.java
noahheller: changed to concurrent hashmap

src/main/java/frc/robot/subsystems/Vision.java
noahheller: done

src/main/java/frc/robot/subsystems/SwerveDrivetrain.java
noahheller: done

PR #189, opened 2024-03-26
Nh gyro thread

noahheller:


src/main/java/frc/robot/utils/ThreadedGyro.java
```diff
@@ -0,0 +1,59 @@
...
+    private boolean enabled = false;
+    private final ScheduledExecutorService executor;
+
+    public ThreadedGyro(AHRS gyro) {
+        this.gyro = gyro;
+        this.angleAdjustment = gyro.getAngleAdjustment();
+        this.executor = Executors.newScheduledThreadPool(1);
+    }
+
+    public void execute(){
```
armadilloz: This method should probably be named `start()`

src/main/java/frc/robot/utils/ThreadedGyro.java
```diff
@@ -0,0 +1,59 @@
...
+        executor.scheduleAtFixedRate(() -> {
+            gyro.setAngleAdjustment(getAngleAdjustment());
```
armadilloz: This should probably not be done on every cycle.

src/main/java/frc/robot/utils/ThreadedGyro.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class ThreadedGyro {
+    private final AHRS gyro;
+    private final AtomicBoolean shouldReset = new AtomicBoolean(false);
+    private volatile double lastGyro;
```
armadilloz: Every variable that is shared between threads should be `Atomic*`

src/main/java/frc/robot/utils/ThreadedGyro.java
```diff
@@ -0,0 +1,59 @@
...
+    private void updateGyro() {
+        lastGyro = (gyro.getAngle() % 360)  * -1;
+    }
+
+    public double getGyroValue(){
+        return lastGyro;
+    }
+
+    public void updateIfDisabled() {
+        if (!enabled){
```
armadilloz: `enabled` is always `false`

src/main/java/frc/robot/utils/ThreadedGyro.java
```diff
@@ -0,0 +1,59 @@
...
+            updateGyro();
```
armadilloz: This is also updating the gyro from another thread, which is not advisable unless AHRS is thread safe

src/main/java/frc/robot/utils/ThreadedGyro.java
```diff
@@ -0,0 +1,59 @@
...
+        }
+    }
+
+    public void resetGyro() {
+        shouldReset.set(true);
+    }
+
+    public void setAngleAdjustment(double degrees) {
```
armadilloz, line 52: This is public but not called from outside

src/main/java/frc/robot/subsystems/SwerveDrivetrain.java
```diff
@@ -127,7 +116,8 @@ public SwerveDrivetrain(SwerveIdConfig frontLeftConfig, SwerveIdConfig frontRigh
         this.frontLeft.getSwerveMotor().getDriveMotor().setInverted(Constants.SWERVE_MODULE_PROFILE.isFrontLeftInverted());
         this.backRight.getSwerveMotor().getDriveMotor().setInverted(Constants.SWERVE_MODULE_PROFILE.isBackRightInverted());
         this.backLeft.getSwerveMotor().getDriveMotor().setInverted(Constants.SWERVE_MODULE_PROFILE.isBackLeftInverted());
-        this.poseEstimator = new SwervePosEstimator(encodedSwerveSparkMaxFL,encodedSwerveSparkMaxFR,encodedSwerveSparkMaxBL,encodedSwerveSparkMaxBR,kinematics,getGyro());
+        gyro.updateIfDisabled();
```
armadilloz: Maybe not the best idea - we can add "synchronousGet" to make sure we have a reading is necessary.

armadilloz (dismissed):

src/main/java/frc/robot/utils/ThreadedGyro.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public void execute(){
+        executor.scheduleAtFixedRate(() -> {
+            gyro.setAngleAdjustment(getAngleAdjustment());
+            if (shouldReset.get()) {
+                gyro.reset();
+                shouldReset.set(false);
+            }
+            updateGyro();
+        },0,10, TimeUnit.MILLISECONDS);
```
armadilloz: That should be a constant and probably larger (20-30 ms)

src/main/java/frc/robot/utils/ThreadedGyro.java
noahheller: done

src/main/java/frc/robot/utils/ThreadedGyro.java
noahheller: I think volatile would have worked because the each thread is either reading or writing (we dont have any operations that both read and write at same time like i++) but it does not hurt. I am now using an atomicLong to represent the byte value of the double because there is no such thing as an atomic double/float. Also, I did not want to use an atomic reference because if for some reason we want to compare to values and use the methods in atomic reference it would compare equality not identity (I think).

src/main/java/frc/robot/utils/ThreadedGyro.java
noahheller: removed, not needed

src/main/java/frc/robot/utils/ThreadedGyro.java
noahheller: The gyro is _technicaly_ somewhat thread safe but I dont want to rely on it. I removed the method it was not needed

src/main/java/frc/robot/utils/ThreadedGyro.java
noahheller: Called by drivetrain subsystem which is called by SetInitOdom Command

src/main/java/frc/robot/subsystems/SwerveDrivetrain.java
noahheller: now I just get the first reading in the constructor of the Gyro. This is not an important value anywasy because we reset pose anyways.

src/main/java/frc/robot/utils/ThreadedGyro.java
noahheller: Made it a constant. In terms of speed Bat-Chen mentioned making it 10 mills so we always have a fresh value for the next tic. I am not opposed to making it 20ms but problemby don't want to go higher or we might have pid issues

bengold22 (approved):

src/main/java/frc/robot/utils/ThreadedGyro.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public void start(){
+        executor.scheduleAtFixedRate(() -> {
+            if (shouldReset.get()) {
+                gyro.reset();
+                shouldReset.set(false);
+            }
+            if (shouldOffset.get()){
+                gyro.setAngleAdjustment(Double.longBitsToDouble(gyroOffset.get()));
+                shouldReset.set(false);
```
armadilloz, line 34: `shouldOffset.set`

src/main/java/frc/robot/utils/ThreadedGyro.java
```diff
@@ -0,0 +1,56 @@
...
+    private final ScheduledExecutorService executor;
+
+    public ThreadedGyro(AHRS gyro) {
+        this.gyro = gyro;
+        this.lastGyro = new AtomicLong((Double.doubleToLongBits(0)));
+        this.executor = Executors.newScheduledThreadPool(1);
+    }
+
+    public void start(){
+        executor.scheduleAtFixedRate(() -> {
```
armadilloz, line 27: We discussed getting the value synchronously prior to starting the thread to ensure a valid reading (reminder)

noahheller: done (in new PR)

src/main/java/frc/robot/utils/ThreadedGyro.java
noahheller: kind of defeats the whole purpose if I don't set it to false idk how i missed that in new PR

PR #196, opened 2024-04-03
Nh multi thread odom

noahheller:


noahheller:
needs to be tested. Work in progress

src/main/java/frc/robot/Robot.java
```diff
@@ -33,6 +37,11 @@ public class Robot extends TimedRobot {
 
     private RobotContainer robotContainer;
     private Command autoCommand;
+    private static final AtomicReference<RobotMode> mode = new AtomicReference<>(RobotMode.DISABLED);
+
+    public static RobotMode getMode(){
```
armadilloz, line 42: Can't this be answered by `Robot.isAutonomous()` etc.?

src/main/java/frc/robot/swervev2/SwervePosEstimator.java
```diff
@@ -58,44 +76,132 @@ public SwervePosEstimator(GenericEncodedSwerve frontLeftMotor, GenericEncodedSwe
                 new Pose2d(),
                 stateStdDevs,
                 visionMeasurementStdDevs);
+        estimatedPose = new AtomicReference<>(poseEstimator.getEstimatedPosition());
         NetworkTableInstance inst = NetworkTableInstance.getDefault();
         NetworkTable table = inst.getTable("ROS");
-        subscriber = table.getDoubleArrayTopic("Pos").subscribe(new double[]{-1,-1,-1});
+        visionMeasurementSubscriber = table.getDoubleArrayTopic("Pos").subscribe(new double[]{-1,-1,-1,-1}, PubSubOption.pollStorage(10), PubSubOption.sendAll(true));
```
armadilloz, line 78: I'm not sure what the impact of the `sendAll` option is going to be for the CPU/network usage. Why not just use the `storage(10)` by itself? Are we losing measurements?

src/main/java/frc/robot/swervev2/SwervePosEstimator.java
```diff
@@ -58,44 +76,132 @@ public SwervePosEstimator(GenericEncodedSwerve frontLeftMotor, GenericEncodedSwe
...
+        apriltagIdSubscriber = table.getIntegerArrayTopic("apriltag_id").subscribe(new long[]{-1,-1});
         SmartDashboard.putData(field);
+        ScheduledExecutorService visionService = Executors.newScheduledThreadPool(2);
+        visionService.scheduleAtFixedRate(() -> {
```
armadilloz: I don't think there is a need for a separate thread here. The updating of the odometry is done at the same rate as the main loop since you're getting all the measurements anyways (and the math calculations are probably fast).
The calculation itself is two stage where the second one is depending on the first one being complete so can definitely be done in a subsequent block once the first is done.
Non-threaded model will simplify the code and remove the need for concurrent data structures which are expensive.

src/main/java/frc/robot/swervev2/SwervePosEstimator.java
```diff
@@ -7,40 +7,54 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * Class to estimate the current position of the robot,
  * given the serve wheel encoders (through swerve model states)
  * and the initial position of the robot
  */
-public class SwervePosEstimator{
+public class SwervePosEstimator {
```
armadilloz, line 37: Wow so this got simplified while I was reviewing :-)

src/main/java/frc/robot/constants/GameConstants.java
```diff
@@ -160,6 +160,9 @@ public class GameConstants {
     public static final double RAMP_AT_POS_THRESHOLD = 0.1;
 
     public static final long GYRO_THREAD_RATE_MS = 10;
+    public static final long POSE_BUFFER_STORAGE_TIME = 2;
+    public static final boolean MULTI_CAMERA = false;
```
armadilloz, line 164: This is not used.

src/main/java/frc/robot/swervev2/SwervePosEstimator.java
```diff
@@ -7,40 +7,54 @@
...
 
     /* standard deviation of robot states, the lower the numbers arm, the more we trust odometry */
-    private static final Vector<N3> stateStdDevs = VecBuilder.fill(0.01, 0.01, 0.001);
+    private static final Vector<N3> stateStdDevs = VecBuilder.fill(0.15, 0.15, 0.001);
 
     /* standard deviation of vision readings, the lower the numbers arm, the more we trust vision */
-    private static final Vector<N3> visionMeasurementStdDevs = VecBuilder.fill(0.7, 0.7, 0.3);
+    private static final Vector<N3> visionMeasurementStdDevs = VecBuilder.fill(0.4, 0.4, 0.5);
+    private static final Transform2d cameraOneTransform = new Transform2d(Constants.CAMERA_OFFSET_FROM_CENTER_X, Constants.CAMERA_OFFSET_FROM_CENTER_Y, new Rotation2d());
+    private static final Transform2d cameraTwoTransform = new Transform2d(Constants.CAMERA_OFFSET_FROM_CENTER_X, Constants.CAMERA_OFFSET_FROM_CENTER_Y, new Rotation2d());
```
armadilloz, line 54: Why do we need `cameraTwoTransform`? It is the same as `cameraOneTransform`.

src/main/java/frc/robot/swervev2/SwervePosEstimator.java
```diff
@@ -7,40 +7,54 @@
...
+    private final AtomicReference<Pose2d> estimatedPose;
```
armadilloz, line 55: This can be made into a simple reference now.

src/main/java/frc/robot/swervev2/SwervePosEstimator.java
```diff
@@ -7,40 +7,54 @@
...
+    private final TimeInterpolatableBuffer<Pose2d> robotPoses = TimeInterpolatableBuffer.createBuffer(Constants.POSE_BUFFER_STORAGE_TIME);
+    private final ConcurrentLinkedQueue<VisionMeasurement> visionPoses = new ConcurrentLinkedQueue<>();
```
armadilloz, line 57: And this can probably be made into a `DEQueue`

src/main/java/frc/robot/swervev2/SwervePosEstimator.java
```diff
@@ -58,44 +72,96 @@ public SwervePosEstimator(GenericEncodedSwerve frontLeftMotor, GenericEncodedSwe
...
     public void updatePosition(double gyroValueDeg){
-        if (DriverStation.isEnabled()){
-            poseEstimator.update(new Rotation2d(Math.toRadians(gyroValueDeg)),
-                    new SwerveModulePosition[] {
-                            frontLeftMotor.getPosition(),
-                            frontRightMotor.getPosition(),
-                            backLeftMotor.getPosition(),
-                            backRightMotor.getPosition(),
+        if (Robot.getMode().equals(RobotMode.TELEOP) && Constants.ENABLE_VISION){
+            long[] tagData = apriltagIdSubscriber.get();
```
armadilloz, line 90: Ideally, you'd like to get the list of April tags and the list of poses since the April tags can change over time. This way you only get the latest.

armadilloz: Alternatively, we can get a few and see if they changed.

Thinking about this some more - this is probably OK as-is.

src/main/java/frc/robot/swervev2/SwervePosEstimator.java
```diff
@@ -58,44 +72,96 @@ public SwervePosEstimator(GenericEncodedSwerve frontLeftMotor, GenericEncodedSwe
...
+            TimestampedDoubleArray[] queue = visionMeasurementSubscriber.readQueue();
+            for (TimestampedDoubleArray measurement : queue){
+                if (validAprilTagPose(measurement)){
+                    double latencyInSec = PrecisionTime.MICROSECONDS.convert(PrecisionTime.SECONDS, measurement.serverTime - TimeUnit.MILLISECONDS.toMicros((long) measurement.value[3]));
+                    visionPoses.add(new VisionMeasurement(measurement, Apriltag.of((int) tagData[0]), latencyInSec));
+                }
+            }
+            while (visionPoses.size() >= 2){
```
armadilloz, line 98: Using measurements while the queue has more than 1 may not use the lates measurement: For example if the queue has 3 entries, this loop will use the older 2 and leave the newest one till next time.
This loop can be made more accurate by using a queue bound to size of 2, where FIFO pushes to it will discard older measurements leaving just the newest 2, and so if the queue has 2 measurements, take them and use them. If there is only one, do nothing and wait till the next cycle.

armadilloz: This will ensure you have used everything you have in each iteration.

src/main/java/frc/robot/swervev2/SwervePosEstimator.java
```diff
@@ -58,44 +72,96 @@ public SwervePosEstimator(GenericEncodedSwerve frontLeftMotor, GenericEncodedSwe
...
+    }
+
+    private Pose2d getVisionPose(TimestampedDoubleArray measurement, Apriltag tag){
+        Translation2d visionPose = new Translation2d(measurement.value[0], measurement.value[1]);
+        Pose2d estPose = estimatedPose.get();
+        double slope = PoseUtils.slope(tag.getPose().toTranslation2d(),new Translation2d(visionPose.getX(), visionPose.getY()));
+        Rotation2d facingAngle = new Rotation2d(Math.atan(slope));
+        Transform2d camTransform;
+        double visionCentricAngle = (estPose.getRotation().getDegrees() + Math.PI) % 360;
+        //this would be used if we had another camera that was mounted 180 degrees from current camera
```
armadilloz, line 148: I wouldn't use that right now - the cameras will have different transformations for them at the ROS level and so this will likely change.
SIMPLIFY :-)

armadilloz: (this probably means you don't need to send the tag in either)

src/main/java/frc/robot/utils/PrecisionTime.java
```diff
@@ -0,0 +1,31 @@
...
+    private static final long NANO_SCALE   = 1L;
+    private static final long MICRO_SCALE  = 1000L * NANO_SCALE;
+    private static final long MILLI_SCALE  = 1000L * MICRO_SCALE;
+    private static final long SECOND_SCALE = 1000L * MILLI_SCALE;
+    private final long scale;
+
+    PrecisionTime(long scale) {
+        this.scale = scale;
+    }
+    public double convert(PrecisionTime dest, double value){
```
armadilloz, line 26: Looks like we need time units in double seconds (which is an odd choice, but oh well, WPILIB requires it). Better to follow the pattern used by `Duration` API since it is pretty much the standard.

src/main/java/frc/robot/utils/math/ArrayUtils.java
```diff
@@ -0,0 +1,52 @@
+package frc.robot.utils.math;
+
+public class ArrayUtils {
```
armadilloz, line 3: Seems like `Arrays` support at least some of these operations.

armadilloz (changes requested):
Thanks for working on that so quickly!
My biggest comment from the earlier version is that the threading model was not right (and therefore some of the implementation was not right). This looks much better.
I would not suggest you make all the changes proposed here in light of the timeline and the looming competition. Since the current code is tested, let's use that and then improve on it after the competition.

src/main/java/frc/robot/Robot.java
noahheller: a lot of people were reporting that it takes a lot of time and can cause CommandLoopOverruns because it has to request the information from the driver station itself

src/main/java/frc/robot/swervev2/SwervePosEstimator.java
noahheller: sendAll with queue 20 messages, there is no reason that we would ever have more than 10. In reality, it does not matter because we read the messages quick enough that it won't even get to 10. I am not sure about the functionality of just having pollStorage we would have to test it. (When I coded this I did not know how many vision measurement we were getting in one tick

src/main/java/frc/robot/swervev2/SwervePosEstimator.java
noahheller: the current code also doesnt even use the apriltag Id it results in the same value being returned from the getVisionPose function. (only used if we have two camera mounted the two transforms are set to be the same)

src/main/java/frc/robot/swervev2/SwervePosEstimator.java
noahheller: Do we want to discard the old measurements? Is it a problem to just use it in the next tic?

src/main/java/frc/robot/swervev2/SwervePosEstimator.java
noahheller: I agree. Hopefully for worlds it will happen at the ros level.

src/main/java/frc/robot/utils/math/ArrayUtils.java
noahheller: Would this be the equivalent of allMatch?
```
double[] comp = new double[measurement.value.length];
Arrays.fill(comp,-1);
Arrays.equals(measurement.value,comp);
```
Or is there a another method without streams?

src/main/java/frc/robot/utils/PrecisionTime.java
noahheller: I was struggling to find the build in java version because they all seem to return longs so they dont have precision meaning I cant get fractional seconds.

src/main/java/frc/robot/swervev2/SwervePosEstimator.java
noahheller: it is the same because this code was going to account for two camera but we did not get there (also we will problably use ros for the offset in the future)

src/main/java/frc/robot/swervev2/SwervePosEstimator.java
noahheller: TODO: remove thread safe datatypes (only using main thread)

